PR #4, opened 2017-10-29
Some fixes

ozancaglayan:
Hello.

Now it's working :)


mjpost:
Thanks for this. Have you run test.sh to make sure the scores are +- 0.4 with Moses' mteval-v13a.pl? (I suppose I should cache its output so as to remove the Moses dependency from the test script).

I'm away from my computer this week but will look into this when I get back.

PR #11, opened 2017-11-07
Add bootstrap resampling

cfedermann:
This includes the SIGPIPE fix. Call with `--bootstrap-trials $n` or `-b $n`. Any `n>1` will result in bootstrap resampling to determine the BLEU score. If `numpy` is available, the code also computes the 0.95 confidence interval around the final BLEU score.

Uses a fixed random seed `12345` to guarantee reproducible scores. This could later be made configurable in which case the sacreBLEU signature needs to be updated. Not needed for now, though.

### DIFF
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -673,25 +673,22 @@ def download_test_set(test_set, langpair=None):
 BLEU = namedtuple('BLEU', 'score, ngram1, ngram2, ngram3, ngram4, bp, sys_len, ref_len')
 
 
-def compute_bleu(instream, refstreams, smooth=0., force=False, lc=False, tokenize=False) -> BLEU:
+def compute_bleu(instream, refstreams, smooth=0., force=False, lc=False, tokenize=False, bootstrap_trials=1) -> BLEU:
     """Produces the BLEU scores along with its sufficient statistics from a source against one or more references.
 
     :param instream: the input stream, one segment per line
     :param refstreams: a list of reference streams
+    :param bootstrap_trials=1: number of trials for bootstrap resampling
     :return: a BLEU object containing everything you'd want
     """
 
     fhs = [sys.stdin] + refstreams
 
-    sys_len = 0
-    ref_len = 0
-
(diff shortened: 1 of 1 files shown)


mjpost:
Thanks for this! I'd like to simply modify the one-line output for computer-readability, i.e.,

    cat newstest2017.uedin-nmt.4955.cs-en | ./sacrebleu.py -t wmt17 -l cs-en -b 10
Mean BLEU score: 30.99 +/- 0.20
    BLEU+case.mixed+lang.cs-en+numrefs.1+test.wmt17+tok.13a+version.1.0.4 = 30.99 +/- 0.20 n=10 62.4/36.9/24.4/16.4 (BP = 1.000 ratio = 1.004 hyp_len = 61946 ref_len = 61718)

Any objections?

cfedermann:
Perfectly fine, go ahead :)

mjpost:
So, don't hate me, but I had to refactor the main class to make an API. Do you want to try to rebase off master? If you don't get to it today I'll do that next. Then I'll push this out as version 1.1.0 and can hopefully leave it alone for a while...

mjpost:
Another issue: have you tested this against the Moses implementation to ensure the results are the same?

cfedermann:
I'll look into this shortly.

ozancaglayan:
Hello,

There's this `random.statistics` standard library in Python >= 3.4 so for mean and stdev you may want to drop numpy dependency if @mjpost is OK to define a minimum dep of 3.4.

martinpopel:
As for the numpy dependency: despite the name sacreBLEU it would be nice to add also a character-based metric, e.g. chrF3 and there is a numpy implementation (probably much faster than the original pure-Python): https://github.com/awslabs/sockeye/pull/216/files

mjpost:
I'm going to close this now in light of its age. Please feel free to re-open it if you have the time and inclination!

mjpost:
Want to pick this up again, @cfedermann?

PR #30, opened 2019-05-23
BLEU obj to string format, and typing for corpus_bleu

thammegowda:
Embedding this directly into my python program (thanks for making this library!). 
I made some minor improvements, hoping that others may like to have it as part of the library.

- BLEU string formatting is now part of BLEU object itself. (so we can log/print it easily) 

- added type annotation to corpus_bleu args. common pitfall (which I also fell) when working with single references is ref_stream assumed to be Iterable[str] instead of List[Iterable[str]]

(tested on python 3.6, it produces the same string as before.)

### DIFF
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -36,7 +36,7 @@ import urllib.request
 
 from collections import Counter, namedtuple
 from itertools import zip_longest
-from typing import List, Iterable, Tuple
+from typing import List, Iterable, Tuple, Union
 
 VERSION = '1.3.3'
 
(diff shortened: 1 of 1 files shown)


mjpost:
Thanks!

francoishernandez:
Hey,
FYI this breaks compatibility with python 3.5. 
`precisions = "/".join(f"{p:.1f}" for p in self.precisions)`
Something a bit heavier but working:
`precisions = "/".join(["{:.1f}".format(p) for p in self.precisions])`
Not sure you care but wanted to mention it just in case.

mjpost:
Sure, there’s no reason to use f-strings. Do you want to submit a PR?

PR #35, opened 2019-06-25
Added file locking on download (closes #31)

### LINKED ISSUE #31: Running two or more instances of sacrebleu in a pipe does not lock temp files
Hi,
When I run a command where sacrebleu is used multiple times like this (I do this all the time):

`./sacrebleu.py -t wmt16 -l ende --echo src | ./generic_translator | ./sacrebleu.py -t wmt16 -l ende`

the two instances may overwrite each others temporary files when the test set is being downloaded for the first time. Both instances then start downloading, unpacking, etc. They may end-up corrupting the files.

mjpost:


mjpost:
This was merged in commit 138f3f658bebaa1855a4af4eb43935bd9bd7c00c. Unfortunately, I stuck with `portalocker`, since it was easier.

PR #41, opened 2019-08-15
Tiny improvements, nicer "sacrebleu -h"

martinpopel:
See the logs of individual commits for details and motivations.
None of these changes affects the sacreBLEU behavior.
I cherry-picked these (hopefully non-controversial) commits from another branch where I am adding new features (which I plan to put in a separate PR soon, ideally after this PR is merged).

PR #42, opened 2019-08-16
Allow evaluation on concatenation of multiple test sets

martinpopel:
This PR is just about the last commit (2c0e788), the previous commits are part of https://github.com/mjpost/sacreBLEU/pull/41 (and should be discussed there).

I used bash for the test to keep consistency with the existing test.sh and also because I can test few other things (e.g. #31) more easily than with pytest.
I think @cfedermann is considering to rewrite all the tests into Python.

### DIFF
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1599,8 +1599,6 @@ def main():
 
     # concat_ref_files is a list of list of reference filenames, for example:
     # concat_ref_files = [[testset1_refA, testset1_refB], [testset2_refA, testset2_refB]]
-    # Note that concatenation is possible only for internal test sets,
-    # which do not currently support multiple references, so the example is hypothetical.
     if args.test_set is None:
         concat_ref_files = [args.refs]
     else:
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1553,11 +1553,19 @@ def main():
         logging.error('Unknown test set "%s"\n%s', args.test_set, get_a_list_of_testset_names())
         sys.exit(1)
 
-    if args.test_set and (args.langpair is None or args.langpair not in DATASETS[args.test_set]):
-        if args.langpair is None:
-            logging.error('I need a language pair (-l).')
-        elif args.langpair not in DATASETS[args.test_set]:
-            logging.error('No such language pair "%s"', args.langpair)
+    if args.test_set is None:
+        if len(args.refs) == 0:
+            logging.error('I need either a predefined test set (-t) or a list of references')
+            logging.error(get_a_list_of_testset_names())
+            sys.exit(1)
+    elif len(args.refs) > 0:
+        logging.error('I need exactly one of (a) a predefined test set (-t) or (b) a list of references')
+        sys.exit(1)
+    elif args.langpair is None:
+        logging.error('I need a language pair (-l).')
+        sys.exit(1)
+    elif args.langpair not in DATASETS[args.test_set]:
+        logging.error('No such language pair "%s"', args.langpair)
         logging.error('Available language pairs for test set "%s": %s', args.test_set,
                       ', '.join(filter(lambda x: '-' in x, DATASETS[args.test_set].keys())))
         sys.exit(1)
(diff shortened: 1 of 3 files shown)


mjpost:
This looks very nice. Can you merge on master so I can see the changes better here?

martinpopel:
Anything else I should do?

I tried to separate each PR into several commits which are easier to review and understand their motivation based on the commit log. The later PR's are based on the earlier ones, so there should be no conflicts when merging the oldest PRs first. Alternatively, you can merge just the last PR (#44) which includes all the commits, if you consider everything OK.

sacrebleu.py
```diff
@@ -1549,9 +1549,11 @@ def main():
...
         sys.exit(1)
 
-    if args.test_set is not None and args.test_set not in DATASETS:
-        logging.error('Unknown test set "%s"\n%s', args.test_set, get_a_list_of_testset_names())
-        sys.exit(1)
+    if args.test_set is not None:
+        for test_set in args.test_set.split(','):
+            if test_set not in DATASETS:
+                logging.error('Unknown test set "%s"\n%s', test_set, get_a_list_of_testset_names())
+                sys.exit(1)
```
mjpost: How about adding `nargs='+'` to `args.test_set`, in which case it would already be a list? I think I prefer this, too, since then the test sets can follow the param as a space-delimited list, instead of comma-delimited.

mjpost:
Sorry, just returned from lots of traveling, trying to make my way through things. I'd prefer to do these one by one and remerge on master after each one, if you don't mind. Thanks very much for your patience! I will commit to looking thoroughly at one of these each night this week.

sacrebleu.py
```diff
@@ -1468,7 +1468,7 @@ def main():
                                          #epilog = 'Available test sets: ' + ','.join(sorted(DATASETS.keys(), reverse=True)),
                                          formatter_class=argparse.RawDescriptionHelpFormatter)
     arg_parser.add_argument('--test-set', '-t', type=str, default=None,
-                            help='the test set to use (see also --list)')
+                            help='the test set to use (see also --list) or a comma-separated list of test sets to be concatenated')
     arg_parser.add_argument('-lc', action='store_true', default=False,
```
mjpost: See note below, this could be space-delimited if we use builtin argparse primitives.

sacrebleu.py
```diff
@@ -1536,11 +1536,11 @@ def main():
...
-            logging.error('No citation found for %s', args.test_set)
-            sys.exit(1)
-
-        print(DATASETS[args.test_set]['citation'])
+        for test_set in args.test_set.split(','):
+            if 'citation' not in DATASETS[test_set]:
+                logging.error('No citation found for %s', test_set)
+            else:
+                print(DATASETS[test_set]['citation'])
         sys.exit(0)
```
mjpost: This is good.

sacrebleu.py
```diff
@@ -1592,30 +1597,38 @@ def main():
...
-        if len(ref_files) == 0:
-            print('No references found for test set {}/{}.'.format(args.test_set, args.langpair))
-            sys.exit(1)
+    # concat_ref_files is a list of list of reference filenames, for example:
+    # concat_ref_files = [[testset1_refA, testset1_refB], [testset2_refA, testset2_refB]]
+    # Note that concatenation is possible only for internal test sets,
+    # which do not currently support multiple references, so the example is hypothetical.
+    if args.test_set is None:
+        concat_ref_files = [args.refs]
     else:
```
mjpost: This isn't right—we do have multiple references for some test sets, see `wmt17/tworefs`.

sacrebleu.py
martinpopel: Personally, I prefer comma-delimited over space-delimited for this purpose, but I will change this if you really think it is better. Note that:
* The semantics of `-m bleu chrf` is that we want to do two separate evaluations (and we are lazy to write a bash for cycle). However, the semantics of `-t wmt18,wmt19` is not to report two BLEU scores, but to report a single BLEU on the concatenation. Originally, I chose the plus symbol (`-t wmt18+wmt18`) instead of comma to make this explicit, but it clashes with the plus symbols in signature. Ampersand symbol is not bash friendly, so in the end I chose the comma.
* Some users prefer the style with long option names and the equal symbol (`--option=value`), which is not compatible with `nargs='+'`. Currently, you can write `--test-set=wmt18,wmt19`, but `--test-set=wmt18 wmt19` would not work (similarly as `--metrics=bleu chrf` does not work, which annoys me a bit).
* We probably don't want to use spaces in the signature, so we would have to come up with some symbol anyway. It seems a bit confusing to use comma (or whatever) in the signature, but force users to use space on the command line.

mjpost: Okay, you convinced me on this :)

sacrebleu.py
mjpost: Is there anything to say here? I think this is the only blocker on this one.

martinpopel: You are right, I checked that `-t wmt16/tworefs,wmt17/tworefs -l en-fi` works as expected and deleted the misleading comment.
 (I had thought that multiple refs can be only specified as multiple files on the command line or a single file with tabs, but now I see that it is supported also for the internal test sets.)

PR #43, opened 2019-08-17
Add --origlang

martinpopel:
RFC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
    - All improvements below due to Martin Popel
    - Added evaluation on multiple test sets (e.g., `-t wmt17,wmt18`).
      Works as long as they all have the same language pair.
-   - Added `sacrebleu --origlang` (both for evaluation on a subset and for `--echo`)
+   - Added `sacrebleu --origlang` (both for evaluation on a subset and for `--echo`).
+   - Added `sacrebleu --detail` for breakdown by subsets of the test set.
    - Minor changes
      - Improved display of `sacrebleu -h`
      - Added `sacrebleu --list`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,13 @@
    - All improvements below due to Martin Popel
    - Added evaluation on multiple test sets (e.g., `-t wmt17,wmt18`).
      Works as long as they all have the same language pair.
+   - Added `sacrebleu --origlang` (both for evaluation on a subset and for `--echo`)
    - Minor changes
      - Improved display of `sacrebleu -h`
      - Added `sacrebleu --list`
      - Code refactoring
-     - Documentation updates
+     - Documentation and tests updates
+     - Fixed a race condition bug (`os.makedirs(outdir, exist_ok=True)` instead of `if os.path.exists`)
 
 - 1.3.7 (2019-07-12)
    - Lazy loading of regexes cuts import time from ~1s to nearly nothing (thanks, @louismartin!)
--- a/README.md
+++ b/README.md
@@ -50,6 +50,16 @@ Use `--short` for a shorter hash if you like.
 
     BLEU+case.mixed+lang.de-en+test.wmt17 = 32.97 66.1/40.2/26.6/18.1 (BP = 0.980 ratio = 0.980 hyp_len = 63134 ref_len = 64399)
 
+If you are interested in the translationese effect, you can evaluate BLEU on a subset of sentences
+with a given original language (identified based on the origlang tag in the raw SGM files).
+E.g. to evaluate only against originally German sentences translated to English use:
+
+    sacrebleu -t wmt13 -l de-en --origlang=de < my-wmt13-output.txt
+
+and to evaluate against the complement (in this case origlang en, fr, cs, ru, de) use:
+
+    sacrebleu -t wmt13 -l de-en --origlang=non-de < my-wmt13-output.txt
+
 # MOTIVATION
 
 Comparing BLEU scores is harder than it should be.
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ This installs a shell script, `sacrebleu`.
 
 Get a list of available test sets:
 
-    sacrebleu
+    sacrebleu --list
 
 Download the source for one of the pre-defined test sets:
 
@@ -76,7 +76,7 @@ SacreBLEU is licensed under the Apache 2.0 License.
 
 This was all Rico Sennrich's idea.
 Originally written by Matt Post.
-The official version can be found at <https://github.com/awslabs/sockeye/tree/master/sockeye_contrib/sacrebleu>; the development version at <https://github.com/mjpost/sacrebleu>.
+The official version can be found at <https://github.com/mjpost/sacrebleu>.
 
 If you use SacreBLEU, please cite the following:
 
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1469,11 +1469,11 @@ def get_a_list_of_testset_names():
     return message
 
 
-def _filter_subset(systems, args):
+def _filter_subset(systems, test_sets, langpair, origlang):
     """Filter out sentences with a given origlang according to the raw SGM files."""
     indices_to_keep = []
-    for test_set in args.test_set.split(','):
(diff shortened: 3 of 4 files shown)


mjpost:
Okay, please merge in `master`, update the CHANGELOG, and push up here. I will review this today.

Once these are all merged, I'll push sacrebleu 1.4.0 to pypi.

(BTW, with these external collaborations and dependencies, we should really move to a more extensive `pytest`-based test suite...)

martinpopel:
OK, I merged in master, updated CHANGELOG and also fixed a literal CR in test.sh.
(I would also prefer pytest-based tests, but I kept the current style.
Also, I wanted to test the concurrent sacrebleu processes in pipeline and I don't know how to test this with pytest.)

mjpost:
This is great, thank you! Thanks for iterating on this!

mjpost:
Okay, so I am basically ready to merge this, but I *really* don't like that `--ol` doesn't work during evaluation. This really breaks a common use case. It seems like it should be possible to add this functionality, right?

martinpopel:
I don't understand what do you mean. The main purpose of `--origlang` is to work during evaluation and I hope it does.
The secondary purpose/usage is that it works with `--echo`.

mjpost:
I'm referring to this (last line):

> 1) Generate the src and ref subsets using
>   `sacrebleu -t wmt18 -l en-de --origlang en --echo src > my-orig-en.src`
>   `sacrebleu -t wmt18 -l en-de --origlang en --echo ref > my-orig-en.ref`
>   Translate only the subset.
>   Evaluate it with `sacrebleu my-orig-en.ref < my-orig-en.translation`.
>   Note that there is no --origlang in the evaluation.

e.g., I can't do

```bash
echo output.detok | sacrebleu -t wmt18 -l en-de -ol en
```
and get a score.

martinpopel:
This is a misunderstanding. Your command works.
See the `sacrebleu -t wmt13 -l de-en --origlang=de < my-wmt13-output.txt` example in README.md or the test (https://github.com/martinpopel/sacreBLEU/blob/testset-origlang/test.sh#L42).
If you the input file (`output.detok` or `my-wmt13-output.txt`) is the translation of the whole test set, you must provide the `--origlang` option.

The secondary use case is that you let your MT system translate only the subset of sentences with a given origlang. My comment you cite is only about this second use case.
I don't promote this use case in README, so it does not confuse the users.
But if there is someone who needs this use case (e.g. for tuning hyper-parameters only on a given subset), I added the option to use `--origlang` together with `--echo` end it works as expected.

mjpost:
So in evaluation mode, `-ol` *expects* the full corpus but only *scores* half of it. I think this is inconsistent, and their is potential for confusion because `--echo` returns just the half of the corpus you want.

To put this another way, this should work:

```bash
$ ./sacrebleu.py -t wmt18 -l en-de -ol en --echo src | ./sacrebleu.py -t wmt18 -l en-de -ol en
sacreBLEU: The input and reference stream(s) were of different lengths.
sacreBLEU: 
This could be a problem with your system output or with sacreBLEU's reference database.
If the latter, you can clean out the references cache by typing:

    rm -r /Users/post/.sacrebleu/wmt18

They will be downloaded automatically again the next time you run sacreBLEU.
```

(where in a typical call, you'd have the translation engine between these).

martinpopel:
I admit this may be considered inconsistent, although there was no option for evaluation on subsets yet, so there is no precedent.
I've spent some time thinking about the design and I did a mini poll among my colleagues and this is the best option we were able to find (consensus on). That said, I welcome any better alternative proposals and I am happy to implement them.

My observations (sorry, it's lengthy, but without considering all these aspects we may draw wrong conclusions):
* It is not always "half of the corpus" because older WMT test sets contain 5 different origlangs. This is also the reason why I implemented the "non-" prefix (`--origlang=non-en`).
* In the following PR (#44), I propose `--subset` to constraint the domain (topic) and country of the evaluated documents. There are currently 6 domains and 4 countries, but the option accepts regular expressions.
* I've added also the `--verbose` option (see [an example in test.sh](https://github.com/martinpopel/sacreBLEU/blob/76f137c4443fbe9fc40803ba357a1f9a250cc0aa/test.sh#L47-L60)), which subsumes both `--origlang` and `--subset` and I find it quite useful (adding `--verbose`  was my primary motivation for all the PRs). This option obviously needs the whole test set on input.
* If we abstract from the implementation and API (I mean command line option names and syntax), there are two use cases:
  - *full-input*: Input a translation of the full test set. I think this is the primary use case (based on my mini-poll). This is needed for `--verbose`. It is also needed when evaluating older translations (e.g. downloaded from statmt.org) for which we have no split by origlang. Of course, it is relatively easy to write a script which filters the plain-text translation file using the respective src/ref sgm file, but it is a hassle and the whole idea of sacreBLEU is to prevent such hassles.
  - *filtered-input*: Input a subset of the full test set. Someone may be interested in evaluation on a given origlang/domain only, but multiple times e.g. for hyper-parameter tuning (possibly on multiple wmt years concatenated, to make the filtered test set reasonably big). It is much faster to let the MT system translate only the filtered subset. It would be relatively easy to write a script which inserts empty lines into the translation on the positions which will be ignored with `--origlang`/`--subset`, but again it is a hassle.
* Implementation-wise: Originally, I implemented just the *filtered-input* and cached the filtered referenes in `~/.sacrebleu/$TESTSET/$LANGFILE.orig=$ORIGLANG` (see [my old-origlang branch](https://github.com/mjpost/sacreBLEU/compare/master...martinpopel:old-origlang)). This caching is a bit faster than the current implementation for repeated evaluations, but it does not scale well for `--subset` (and combinations of `--origlang` and `--subset`; it would clutter `~/.sacrebleu/`) and more importantly it does not support the *full-input* use case.

So what are the alternative proposals?
* Forbid the *filtered-input* use case and report an error if anyone tries to use `--origlang` with `--echo`. I'm OK with this if you think it is better.
* In addition to `--origlang`, provide `--origlang-for-the-purpose-of-echo` (and similarly `--subset-for-the-purpose-of-echo`). I don't like this (even if we choose a shorter option name).
* Make the evaluation with `--origlang` auto-magically clever, so that it accepts both the *full-input* and the *filtered-input*. So if the number of lines does not match the full-input and `--origlang` is provided, we can try to filter the reference and if the number of lines matches we will expect the *filtered-input* and print the scores. I don't like auto-magic which tries to pretend simplicity for (new) users but has many negative side effects. But if there is a consensus that this is the best option...
* Any other proposal?

mjpost:
Thanks for this, Martin! I'll do my best to get to this today!

mjpost:
Can you merge in and push up the latest master to this branch, so I can better see the current diff?

mjpost:
Here is my current thinking based on my current understanding (which might be a bit off based on the difficulty of seeing the actual diff):

- I think the full-input scenario makes the most sense. The decoder has to translate the full test set, and at evaluation time, the evaluator (via command-line flags) can do what it likes (including ignoring some, in the case of `--origlang`). I agree that doing something automagically on the eval side is not a good idea. I would like users to know what they're doing.

- I had originally protested that `--echo --origlang` gave a subset (I realize it's not exactly half) but eval expected the whole. I think this is okay. If the user *really* wants to avoid translating part of the test set that won't get evaluated, they could `--echo both` and then dump the reference to file and evaluate with it directly.

- I don't like the flag `--verbose`. That flag to me is a very general flag across all commands in UNIX-space that means something like "print some more information about what you're doing to the output", i.e., it's orthogonal to what you're having it do. I think something like `--detail` or `--breakdown` makes more sense for the use case you're discussing. For example, there might be a reason you'd want to have a *verbose* domain breakdown (instead of just a regular domain breakdown).

- Can you confirm that the (sub-)domain information you are printing comes from the XML files? Or is it annotations you yourself have made?

martinpopel:
Master is already merged into this branch (and https://github.com/mjpost/sacreBLEU/pull/43/files shows the actual diff of this PR).

I renamed `--verbose` to `--detail` and I agree it is a better name, especially if you plan to add `--verbose` with different semantics in future.

The domain information is not part of this PR, but of the following one (#44). Unlike origlang, the domain information is not included in the original SGM files. I added the (stand-off) annotation directly to the source code for reasons described in the PR #44.

PR #44, opened 2019-08-22
Add --subset option and domain+country annotation for wmt18+wmt19

martinpopel:
There are surprising differences in BLEU scores between different domains (topics),
but most researchers ignore this because there is no simple way how to analyze it.
Now it is as simple as adding `--detail` to the `sacrebleu` command (see test.sh).

In addition to 6 domains: business, crime, entertainment, politics, scitech, sport, world,
this PR adds metadata about the main country/region for each news article:
EU, GB, US, OTHER (OTHER is further subclassified, but this annotation is not shown in `--detail`).
This annotation can be useful for disentangling the effect of translationese (original language) and domain/topics.
For such analysis, it would be nice to add domain+country annotation for some originally non-English documents.

I've tried to implement (and name) the `--subset` option in a relatively versatile way,
so it can be used for other purposes in future if someone adds other annotation.
Versioning the annotation (document metadata missing in the original sgm files)
in the source code is not elegant, but it is practical because
- sacrebleu.py can still be used without installing (if portalocker is installed)
- the annotation is tied with SacreBLEU's version
- it is just 10-20 lines of "code" per wmt testset (original English docs only).

### DIFF
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1790,9 +1790,12 @@ def main():
 
     if args.verbose:
         sents_digits = len(str(len(full_system)))
-        for origlang in _available_origlangs(args.test_set, args.langpair):
+        origlangs = args.origlang if args.origlang else _available_origlangs(args.test_set, args.langpair)
+        for origlang in origlangs:
             subsets = [None]
-            if any(t in SUBSETS for t in args.test_set.split(',')):
+            if args.subset is not None:
+                subsets += [args.subset]
+            elif all(t in SUBSETS for t in args.test_set.split(',')):
                 subsets += COUNTRIES + DOMAINS
             for subset in subsets:
                 system, *refs = _filter_subset([full_system, *full_refs], args.test_set, args.langpair, origlang, subset)
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1469,11 +1469,11 @@ def get_a_list_of_testset_names():
     return message
 
 
-def _filter_subset(systems, args):
+def _filter_subset(systems, test_sets, langpair, origlang):
     """Filter out sentences with a given origlang according to the raw SGM files."""
     indices_to_keep = []
-    for test_set in args.test_set.split(','):
-        rawfile = os.path.join(SACREBLEU_DIR, test_set, 'raw', DATASETS[test_set][args.langpair][0])
+    for test_set in test_sets.split(','):
+        rawfile = os.path.join(SACREBLEU_DIR, test_set, 'raw', DATASETS[test_set][langpair][0])
         if not rawfile.endswith('.sgm'):
             raise Exception('--origlang supports only *.sgm files, not %s', rawfile)
         number_sentences_included = 0
@@ -1482,15 +1482,15 @@ def _filter_subset(systems, args):
             for line in fin:
                 if line.startswith('<doc '):
                     doc_origlang = re.sub(r'.* origlang="([^"]+)".*\n', '\\1', line)
-                    if args.origlang.startswith('non-'):
-                        include_doc = doc_origlang != args.origlang[4:]
+                    if origlang.startswith('non-'):
+                        include_doc = doc_origlang != origlang[4:]
                     else:
-                        include_doc = doc_origlang == args.origlang
+                        include_doc = doc_origlang == origlang
                 if line.startswith('<seg '):
                     indices_to_keep.append(include_doc)
                     number_sentences_included += 1 if include_doc else 0
         if number_sentences_included == 0:
-            logging.warning("Test set %s contains no sentence with origlang=%s", test_set, args.origlang)
+            logging.warning("Test set %s contains no sentence with origlang=%s", test_set, origlang)
     return [[sentence for sentence,keep in zip(sys, indices_to_keep) if keep] for sys in systems]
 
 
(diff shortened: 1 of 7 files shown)


sacrebleu.py
```diff
@@ -656,6 +656,46 @@
...
+        'nytimes.184830=US-politics scotsman.133801=GB-sport cnbc.com.6770=US-business bbc.381760=GB-crime reuters.218865=OTHER-world-ID newsweek.51339=US-crime '
+        'euronews-en.153797=OTHER-world-ID abcnews.306774=US-crime dailymail.co.uk.298696=GB-politics abcnews.306755=US-politics reuters.218909=US-crime '
+        'independent.226349=OTHER-sport-RU newsweek.51330=US-politics bbc.381705=GB-sport newsweek.51340=OTHER-world-ID cbsnews.248411=OTHER-world-FM abcnews.306776=US-crime '
+        'bbc.381694=GB-entertainment rt.com.91356=US-world telegraph.405430=GB-entertainment telegraph.405404=EU-world bbc.381749=GB-world telegraph.405413=US-politics '
+        'bbc.381736=OTHER-politics-KP cbsnews.248394=US-politics nytimes.184822=US-world telegraph.405408=US-politics euronews-en.153799=OTHER-politics-SY '
+        'euronews-en.153826=EU-sport cnn.304400=US-world'
+}
+SUBSETS = {k:{d.split('=')[0]:d.split('=')[1] for d in v.split()} for (k,v) in SUBSETS.items()}
+COUNTRIES = sorted(list({v.split('-')[0] for v in SUBSETS['wmt19'].values()}))
+DOMAINS = sorted(list({v.split('-')[1] for v in SUBSETS['wmt19'].values()}))
```
mjpost: I think this would be a lot more readable (and recreatable) if you encoded the dictionary format directly, e.g.,

```python
'wmt19': {
  'GB-politics': ['bbc.381790', 'guardian.221679', '...'],
  '...'
}
```

sacrebleu.py
```diff
@@ -1461,6 +1516,59 @@ def get_a_list_of_testset_names():
...
+                        doc_id = re.sub(r'.* docid="([^"]+)".*\n', '\\1', line)
+                        if not re.search(subset, doc_to_tags.get(doc_id, '')):
+                            include_doc = False
+                if line.startswith('<seg '):
+                    indices_to_keep.append(include_doc)
+                    number_sentences_included += 1 if include_doc else 0
+    return [[sentence for sentence,keep in zip(sys, indices_to_keep) if keep] for sys in systems]
+
+
 def main():
```
mjpost: This would add some work: but I've thought for a while that we should get rid of this hacky XML parsing in favor of using `lxml.etree`. This would simplify a lot of code and remove a potential for errors to creep in. What do you think? We use this library extensively in the Anthology.

sacrebleu.py
martinpopel: The mapping should be docID->annotation, not the opposite way because:
- It is used this way when filtering the test set. I iterate through the test set and check whether a given annotation (e.g `GB-politics`) matches a given **regular expression** specified with `--subset`, so I cannot use mapping you suggest (annotation->list(docID)). I could use it just for specifying the annotation in the source code and build the docID->annotation mapping on the fly, but this seems as an unnecessary obfuscation.
- It is more readable and editable/extensible I think. E.g. it is easy to add more tags to a given docID (the documents are listed in the order they appear in the SGM file).
- My annotation includes also fine-grained country codes, e.g. in `OTHER-politics-MX`, the `MX` means Mexico. So for many annotation strings there would be just a single docID and the source code would be actually longer than now.

While reviewing this code, I admit there is a potential problem that the list of `COUNTRIES` and `DOMAINS` is based on `wmt19` only. Currently `wmt19` countries and domains are a superset of those from `wmt18` and no other test set is annotated, so I opted for keeping the code simple. It can be easily updated when anyone adds annotation for other test-sets with extra domains or countries (although, I consider the current inventory exhaustive as it includes country=OTHER and domain=world, which kind of also means "other").

sacrebleu.py
martinpopel: I would prefer to merge this PR first and possibly do the lxml refactoring in a separate PR, after considering the following:
- I used the ad-hoc reg-ex style, which was already there for the SGM->plaintext conversion. It's just a few lines of code and I doubt the lxml refactoring can be notably shorter/simpler (though I am not familiar with lxml).
- I thought that sacreBLEU is meant to be usable also as a single script with no installation, i.e. no dependencies (meanwhile, `portalocker` was added, I know). That said, if dependencies are allowed now, we could reconsider re-implementing everything with numpy and adding bootstrap resampling (#11) and paired bootstrap tests.
- Are you sure that all legacy WMT *.sgm files are XML 1.0 valid and `lxml.etree` parses them correctly? They have no `<?xml version="1.1" encoding="UTF-8"?>` declaration, so they are not valid XML 1.1. Some older testsets (wmt08, I think) use `<DOC>` instead of `<doc>` etc.

mjpost: I'm not sure, but my impression is that `lxml.etree` is flexible against imperfect specs.

I don't mind if sacrebleu has dependencies, since `pip` makes it easy to install them and doesn't IMO make it harder to use. Rewriting BLEU in `numpy` would be *awesome*. It's too slow as it is. Did you have a reference `numpy` implementation in mind? I hadn't given any thought about how to make this faster.

martinpopel:
Thanks for merging. If time allows, I will try lxml.

Regarding the speed issues, I started answering it here, but then moved it to #46.

mjpost:
Sounds good! I'm going to do some full testing and then I'll release version 1.4.0 to pypi.

tholiao:
Hi Martin, how are the annotations sourced?

PR #45, opened 2019-08-23
Introduce `update_from` param in `corpus_bleu`

francoishernandez:
Hi guys,
I'd like to introduce a way to update some existing BLEU object.
We'd just pass an exisiting `BLEU` object to the `update_from` param of `corpus_bleu`, and it would initialize the values to those of this object instead of 0.
The use case is quite rare but I don't think it hurts to have such an option.

### DIFF
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1275,7 +1275,8 @@ def corpus_bleu(sys_stream: Union[str, Iterable[str]],
                 force=False,
                 lowercase=False,
                 tokenize=DEFAULT_TOKENIZER,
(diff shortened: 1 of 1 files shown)


mjpost:
Hi @francoishernandez, sorry I never got to this. We have since refactored the code so this would have to be updated. I’m happy to support this feature, but I’d prefer to handle it a different way: I think we should factor out the BLEU statistics into a self-contained class, and then overload the `+` operator for it. What do you think?

ozancaglayan:
no answer and the API substantially changed after this. Feel free to re-open if still a required feature.

PR #50, opened 2019-09-29
chrF signature should not include the tokenization variant

martinpopel:
Tokenization is used only for BLEU, not for chrF.
Including `tok.13a` in the chrF signature is thus misleading.
Also, this commit fixes a bug causing `KeyError: 'tok'` when trying to use `-m chrf --short`.


mjpost:
Thank you!

PR #55, opened 2019-10-15
Fixed bug in add-k smoothing

sn1c:
The smoothing value k is only added to the nominator and denominator for n-grams with n > 2 due to an off-by-one error. This is now fixed.

### DIFF
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1609,7 +1609,7 @@ def main():
                             help='Use case-insensitive BLEU (default: actual case)')
     arg_parser.add_argument('--sentence-level', '-sl', action='store_true',
                             help='Output metric on each sentence.')
-    arg_parser.add_argument('--smooth', '-s', choices=['exp', 'floor', 'add-n', 'none'], default='exp',
+    arg_parser.add_argument('--smooth', '-s', choices=['exp', 'floor', 'add-k', 'none'], default='exp',
                             help='smoothing method: exponential decay (default), floor (increment zero counts), add-k (increment num/denom by k for n>1), or none')
     arg_parser.add_argument('--smooth-value', '-sv', type=float, default=SMOOTH_VALUE_DEFAULT,
                             help='The value to pass to the smoothing technique, when relevant. Default: %(default)s.')
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1296,7 +1296,7 @@ def compute_bleu(correct: List[int],
     smooth_mteval = 1.
     effective_order = NGRAM_ORDER
     for n in range(NGRAM_ORDER):
-        if smooth_method == 'add-k' and n > 1:
+        if smooth_method == 'add-k' and n > 0:
             correct[n] += smooth_value
             total[n] += smooth_value
         if total[n] == 0:
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1295,7 +1295,7 @@ def compute_bleu(correct: List[int],
 
     smooth_mteval = 1.
     effective_order = NGRAM_ORDER
-    for n in range(1, NGRAM_ORDER):
+    for n in range(1, NGRAM_ORDER + 1):
         if smooth_method == 'add-k' and n > 1:
             correct[n-1] += 1
             total[n-1] += 1
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1730,6 +1730,9 @@ def main():
             print_test_set(test_set, args.langpair, args.echo, args.origlang, args.subset)
         sys.exit(0)
 
+    if args.smooth == 'add-k' and args.smooth_value == SMOOTH_VALUE_DEFAULT:
+        args.smooth_value = 1
+
     if args.test_set is not None and args.tokenize == 'none':
         logging.warning("You are turning off sacrebleu's internal tokenization ('--tokenize none'), presumably to supply\n"
                         "your own reference tokenization. Published numbers will not be comparable with other papers.\n")
--- a/sacrebleu.py
+++ b/sacrebleu.py
@@ -1295,24 +1295,24 @@ def compute_bleu(correct: List[int],
 
     smooth_mteval = 1.
     effective_order = NGRAM_ORDER
-    for n in range(NGRAM_ORDER):
-        if smooth_method == 'add-k' and n > 0:
-            correct[n] += 1
-            total[n] += 1
-        if total[n] == 0:
+    for n in range(1, NGRAM_ORDER):
+        if smooth_method == 'add-k' and n > 1:
+            correct[n-1] += 1
+            total[n-1] += 1
+        if total[n-1] == 0:
             break
 
         if use_effective_order:
-            effective_order = n + 1
+            effective_order = n
 
-        if correct[n] == 0:
+        if correct[n-1] == 0:
             if smooth_method == 'exp':
                 smooth_mteval *= 2
-                precisions[n] = 100. / (smooth_mteval * total[n])
+                precisions[n-1] = 100. / (smooth_mteval * total[n-1])
             elif smooth_method == 'floor':
-                precisions[n] = 100. * smooth_value / total[n]
+                precisions[n-1] = 100. * smooth_value / total[n-1]
         else:
-            precisions[n] = 100. * correct[n] / total[n]
+            precisions[n-1] = 100. * correct[n-1] / total[n-1]
 
     # If the system guesses no i-grams, 1 <= i <= NGRAM_ORDER, the BLEU score is 0 (technically undefined).
     # This is a problem for sentence-level BLEU or a corpus of short sentences, where systems will get no credit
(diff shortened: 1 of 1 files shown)


mjpost:
Sorry, I haven't had a chance to look at this. Perhaps after ACL...

sacrebleu.py
```diff
@@ -1296,9 +1296,9 @@ def compute_bleu(correct: List[int],
     smooth_mteval = 1.
     effective_order = NGRAM_ORDER
     for n in range(NGRAM_ORDER):
-        if smooth_method == 'add-k' and n > 1:
```
martinpopel: I agree with this bug fix. I suggest to add a comment about 0-based n indexing (but 1-based `NGRAM_ORDER` and `effective_order`).
Alternatively, we could switch to 1-based indexing, i.e. keep `n > 1`, but use `for n in range(1, NGRAM_ORDER+1)` and change all occurrences of indexing by `n` into `n-1`.

sacrebleu.py
```diff
@@ -1296,9 +1296,9 @@ def compute_bleu(correct: List[int],
...
-            correct[n] += smooth_value
-            total[n] += smooth_value
```
martinpopel: This edit is questionable because the comment [add-k: Method 2 (generalizing Lin and Och, 2004)](https://github.com/mjpost/sacreBLEU/blob/f9847a5/sacrebleu.py#L1281) suggests the intent was to generalize add-one smoothing to add-k.

However, I agree the current code is wrong because

* `smooth_value` defaults to `SMOOTH_VALUE_DEFAULT` which is 0.0, so by just using `--smooth=add-k` we don't get any smoothing, which is unexpected for the users.
* There is inconsistency whether `--smooth-value` applies to [`--smooth floor` only](https://github.com/mjpost/sacreBLEU/blob/f9847a5/sacrebleu.py#L73) or anytime [when relevant](https://github.com/mjpost/sacreBLEU/blob/f9847a5/sacrebleu.py#L1615).
* There is inconsistency between `add-k` and `add-n`. Currently, ['exp', 'floor', 'add-n', 'none'](https://github.com/mjpost/sacreBLEU/blob/f9847a5/sacrebleu.py#L1612) are the only allowed values of the `--smooth` parameter, but the rest of the code uses `add-k`, which I consider a better name (because n is used for the n-gram order). Fixing this bug is simple (renaming add-n to add-k), but we can see that no one actually uses this smoothing and it is not covered by the tests. So maybe we could rename it to `add-one` and forget about generalizing which no one uses.

DNGros: +1 to the issue with the --smooth parameter.

I would also adding a extra condition to 
https://github.com/mjpost/sacreBLEU/blob/f9847a52c6b447459a0475a5ae461963ff74ed82/sacrebleu.py#L1308-L1313
```
elif smooth_meth != "none": 
     raise ValueError(f"Invalid smoothing method {smooth_meth}") 
     #                   ^ remove f string python 2 compat
```
This would make sure typos in smooth_meth does not result in 'none' smoothing applied when expect 'add-n' or 'flopr' or whatever...

I also agree that having an explicit "add-one" is option is good option which less likely for the user to mess up by the default smooth-value. Either that or "--smooth-value" should be non-optional when using `--smooth add-k` .

sacrebleu.py
sn1c: I checked the code and saw that the second variant is how it's also done in other locations, e.g. line 1413:
```        for ngram in sys_ngrams.keys():
            n = len(ngram.split())
            correct[n-1] += min(sys_ngrams[ngram], ref_ngrams.get(ngram, 0))
            total[n-1] += sys_ngrams[ngram]```
so I'll change it to 1-based indexing.

File: sacrebleu.py (hunk 2 above)
Comment (sn1c) [reply sacrebleu.py]: @martinpopel You are right. I renamed add-n to add-k and now add-k defaults to 1.
@DNGros I think the check is not necessary as argparse choices parameter restricts the allowed values and after fixing it by renaming add-n to add-k there is no room for typos anymore.

**Comment by sn1c** [comment]
@martinpopel I think I adressed all of your concerns, can you revisit this PR when you are free?

File: sacrebleu.py (hunk 2 above)
Comment (mjpost) [reply sacrebleu.py]: I agree with the move from `add-n` to `add-k`. However, the intent of `--smooth-value` is to be a general value applicable to any smoothing method, and the current changes you proposed hard-code `add-k` to `1`. 

I suggest that we make `SMOOTH_VALUE_DEFAULT` a `dict` indexed by the smoothing methods. e.g.,

```python
SMOOTH_VALUE_DEFAULT = { 'floor': 0.0, 'add-k': 1 }
```
and so on.

DNGros: @sn1c that is a good point. However, it only applies if going through the CLI codepath. If you import sacrebleu and call compute_bleu from python code, I think there is a still a chance it might fail silently due to a typo. This is very minor, but in an ideal world it seems like that should be caught. 

(Don't let this stop this PR from going through though. This is a useful fix which hopefully can be merged soon so people's "add-n"/"add-k" aren't subtly wrong.)

sacrebleu.py
```diff
@@ -1609,7 +1609,7 @@ def main():
                             help='Use case-insensitive BLEU (default: actual case)')
     arg_parser.add_argument('--sentence-level', '-sl', action='store_true',
                             help='Output metric on each sentence.')
-    arg_parser.add_argument('--smooth', '-s', choices=['exp', 'floor', 'add-n', 'none'], default='exp',
+    arg_parser.add_argument('--smooth', '-s', choices=['exp', 'floor', 'add-k', 'none'], default='exp',
                             help='smoothing method: exponential decay (default), floor (increment zero counts), add-k (increment num/denom by k for n>1), or none')
     arg_parser.add_argument('--smooth-value', '-sv', type=float, default=SMOOTH_VALUE_DEFAULT,
                             help='The value to pass to the smoothing technique, when relevant. Default: %(default)s.')
```
DNGros: Consider updating help so that users understand that default is dependent on --smooth

martinpopel:
All my original concerns were addressed, but I have a new one: if `smooth=add-k` is used with a non-default value, this is not reflected in the sacrebleu signature (which indicates only the smoothing method, but not the smoothing value). This means that the signature cannot be trusted and it is better to report the sacrebleu version and command line options (some papers already report the options instead of the signature). However, this bug (if you consider it a bug) is not really related to the current PR and can be fixed in another PR.

The solution with `SMOOTH_VALUE_DEFAULT = { 'floor': 0.0, 'add-k': 1 }` suggested by @mjpost  may be more readable, but it is effectively equivalent to the current solution (because no one should want to use `smooth='add-k', smooth_value=0.0`). As @DNGros pointed out, the code for setting smooth_value could be moved inside the `compute_bleu` method, so it is applied also when using sacrebleu as a library (not as CLI). I don't care much about this.

I suggest to merge this PR soon, so we can move on.

sacrebleu.py
mjpost: I think you've missed the point. If you do a diff from this branch against the current master, you'll see that the value of `smooth_value` is **entirely ignored** for `add-k`. In other words, `add-k` is hard-coded to `k=1`.

mjpost:
Hi, my more immediate concern is that `smooth_value` **is entirely ignored** for `add-k` smoothing.

I agree the smoothing value needs to be added to the signature, too.

So I am paying attention to this now but believe it is not quite ready to merge.

martinpopel:
Oh, yes, `add-k` now ignores the `smooth_value` and this was one of my [original comments](https://github.com/mjpost/sacreBLEU/pull/55/commits/84fba619bec420c259280b8ede9d987f04970fff#r352692068).
So 
* either use the `smooth_value` for `add-k` (with the add-k specific default value 1)
* or rename `add-k` to `add-one`, which would also solve the problem with the signature (well, partially, because there is still the `floor` smoothing).

sn1c:
@mjpost @martinpopel Both of you are totally right! Due to the off-by-one error which caused the tests to fail I completely missed to commit the change to generalize add-k by having it use smooth_value instead of 1. This PR got kinda messy as I opened it some months ago, sorry for that.

martinpopel:
@mjpost: It seems that the `SMOOTH_VALUE_DEFAULT = { 'floor': 0.0, ...}` default won't be intuitive for the users because the floor smoothing (used in `sentence_bleu` by default) with `smooth_value=0` effectively means no smoothing. If no smoothing should be the default for `sentence_bleu`, I would expect explicit `smooth_method='none'` to be more intuitive. See #64.

sacrebleu.py
```diff
@@ -1290,29 +1290,31 @@ def compute_bleu(correct: List[int],
     :param use_effective_order: If true, use the length of `correct` for the n-gram order instead of NGRAM_ORDER.
```
martinpopel: `:param smooth` -> `:param smooth_method`

mjpost:
Hope to look at this on Friday, but it might have to wait until ACL meta-reviews are finished early next week...

mjpost:
Just released [version 1.4.4](https://pypi.org/project/sacrebleu/) with this included. Thank you for your contributions (and patience)!

PR #63, opened 2020-01-10
Close opened tarballs after downloading

tholiao:
Tested the following command:
```sacrebleu --test-set wmt14 --language-pair de-en --echo src > wmt14-de-en.src```


mjpost:
Clearly an improvement, thank you!

Issue #71, opened 2020-04-15
v1.4.6 broke some imports


tholiao:
Previously, I relied on being able to import `DATASETS`, `SACREBLEU_DIR`, `_clean`, and `smart_open` from sacrebleu. The refactor in #66 (9baed3e) broke the latter three imports. 

While I admit that relying on underscore function imports is hacky, I think `SACREBLEU_DIR` is clearly something that should be importable. The case of `smart_open` is more ambiguous: there is a reasonable argument to be made for keeping it "under the hood", and the code is trivial to duplicate, but the backwards incompatibility is both a severe and unexpected change.

mjpost:
Fixed in #029e39edfb733a524d7ea4fc7f64fa93e81b9f53 (1.4.6).

PR #73, opened 2020-04-15
Restore certain imports available in 1.4.5

tholiao:
See #71


tholiao:
Added smart_open to import list. I talked to somebody else at Berkeley whose code broke after smart_open became unavailable to import, so I believe there are probably more people who are hitting the same problem.

mjpost:
Sorry for breaking backwards compatibility! I agree it’s a good idea to export these as well. I’ll bundle up a new release within a day or two. Thanks for the PR.

tholiao:
Thanks @mjpost!

Issue #77, opened 2020-04-29
Remove typing module from dependencies


ErwanDL:
The [external typing module](https://pypi.org/project/typing/) (which is useless now since `typing` has been integrated in the standard library in Python3.5) is being installed with sacrebleu, which causes conflicts with the `typing` module from the std library on recent python versions.

Here is an example of a conflict caused in one of my projects by this : 
```
    File "/root/.cache/pypoetry/virtualenvs/feedly.ml-summarization-py3.7/li  
b/python3.7/site-packages/typing.py", line 1357, in <module>                  
      class Callable(extra=collections_abc.Callable, metaclass=CallableMeta)  
:                                                                             
    File "/root/.cache/pypoetry/virtualenvs/feedly.ml-summarization-py3.7/li  
b/python3.7/site-packages/typing.py", line 1005, in __new__                   
      self._abc_registry = extra._abc_registry                                
  AttributeError: type object 'Callable' has no attribute '_abc_registry'     
```

This dependency is listed in the project's dependencies here :
https://github.com/mjpost/sacrebleu/blob/f54908ac00879f666c92f4174367bcd3a8723197/setup.py#L110

I believe the possible solutions are : 
- conditionally installing the external `typing` module for users with Python < 3.5 only
- removing this dependency altogether, if you do not plan to support Python < 3.5 anyway

Cheers !

mjpost:
I would actually love to support only Python 3.6+, so that I could use f-strings. Maybe I will do this, since people using older versions could always just force-install 1.4.8. Thoughts, @martinpopel?

martinpopel:
I don't have any strong opinion on this, but I like f-strings too.

ErwanDL:
So should I remove this dependency and post a PR ?

mjpost:
Sure, that’d be great. Let me merge #76 first. Then, if you don’t mind, please

- Bump the version to 1.5.0
- Update the CHANGELOG (please add yourself)

mjpost:
Closed in 675a169df04bdab94d68c7607dcd8c03a35b2f92.

PR #81, opened 2020-05-12
Conditionally install typing dependency only for python < 3.5

ErwanDL:
For #77.

I kept the dependency to typing for python < 3.5 instead of removing it altogether as it was easy to do, so that the package will not break for users on old python versions. Consequently, shoud I change the version to 1.4.10 instead of 1.5.0 ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 # VERSION HISTORY
+- 1.5.0 (2020-05-11)
+  - Now only lists the external [typing](https://pypi.org/project/typing/)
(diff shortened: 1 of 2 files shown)


mjpost:
Thanks for this!

Yes, I think maybe do 1.4.10. I was thinking of moving to Python 3.6 so I could use f-strings in the code, but maybe it's better to maintain backwards compatibility. That would require a version bump.

ErwanDL:
I changed to 1.4.10. For the f-strings, maybe moving directly to Python 3.6 is a bit aggressive, but the decision is up to you anyway 😃  !

I think this is ready to merge, thank you !

Issue #5, opened 2017-10-29
Cleaner checking of Chinese charset 


alvations:
Maybe this would be easier to maintain when [checking chinese characterset](https://github.com/mjpost/sacreBLEU/blob/master/sacrebleu.py#L366), the core function don't  change if we were to add more or remove from the charset: 

```
chinese_charset = [
(u'\u3400', u'\u4db5'),  # CJK Unified Ideographs Extension A, release 3.0
(u'\u4e00', u'\u9fa5'),  # CJK Unified Ideographs, release 1.1
(u'\u9fa6', u'\u9fbb'),  # CJK Unified Ideographs, release 4.1
(u'\uf900', u'\ufa2d'),  # CJK Compatibility Ideographs, release 1.1
(u'\ufa30', u'\ufa6a'),  # CJK Compatibility Ideographs, release 3.2
(u'\ufa70', u'\ufad9'),  # CJK Compatibility Ideographs, release 4.1
(u'\u20000', u'\u2a6d6'),  # (UTF16) CJK Unified Ideographs Extension B, release 3.1
(u'\u2f800', u'\u2fa1d'),  # (UTF16) CJK Compatibility Supplement, release 3.1
(u'\uff00', u'\uffef'),  # Full width ASCII, full width of English punctuation, half width Katakana, half wide half width kana, Korean alphabet
(u'\u2e80', u'\u2eff'),  # CJK Radicals Supplement
(u'\u3000', u'\u303f'),  # CJK punctuation mark
(u'\u31c0', u'\u31ef'),  # CJK stroke  
(u'\u2f00', u'\u2fdf'),  # Kangxi Radicals
(u'\u2ff0', u'\u2fff'),  # Chinese character structure
(u'\u3100', u'\u312f'),  # Phonetic symbols
(u'\u31a0', u'\u31bf'),  # Phonetic symbols (Taiwanese and Hakka expansion)
(u'\ufe10', u'\ufe1f'),
(u'\ufe30', u'\ufe4f'),
(u'\u2600', u'\u26ff'),
(u'\u2700', u'\u27bf'),
(u'\u3200', u'\u32ff'),
(u'\u3300', u'\u33ff')
]

def isChineseChar(uchar):
    for start, end in chinese_charset:
        if start <= uchar <= end:
            return turn
    return false
```

alvations:
[...]

Issue #85, opened 2020-05-30
refactor tokenization classes


mjpost:
Tokenization is all hacked together with a dictionary of functions. This has caused problems with the Mecab tokenizer. It would be good to factor this out into a simple class hierarchy.

ozancaglayan:
I started working on this. If you did not do anything so far, you can assign this to me.

mjpost:
Sounds great—thanks, Ozan.

ozancaglayan:
FYI, I also refactored out metrics into their own classes with proper formatters and signature providers. I'll keep pushing onto this branch but I don't know if you'll be willing to accept all of these major refactorings.

ozancaglayan:
Comparing results of sacrebleu and chrF++, I noticed that sacrebleu does not implement the word order parameter and the out-of-the-box numbers produced by the two scripts are rather different unless I pass `-nw 0` to chrF++. Was it OK to implement the metric like this?

m-popovic:
Ah, yes, chrF++ by default takes into account n-grams with 6 characters + 2 words. Pure chrf therefore requires to set -nw 0  (nw = word n-gram order).

ozancaglayan:
Oh so chrF and chrF++ are actually two different metrics then. Thanks!

Issue #27, opened 2019-05-03
(minor) factor tokenizer code


frankseide:
Hi, there are three tokenizers, and at least two of them (`tokenize_13a()` and `tokenize_zh()`) contain the same processing steps for tokenizing Western-script characters. The code is not even identically written, although the regexes seem identical in the end. Code duplication is dangerous, in case future modifications are ever made to this code. I suggest to factor out this code.

mjpost:
It’s a good idea. The main reason I haven’t is because it requires some other work setting up regressions tests and so on to make sure I don’t break anything, but I’ll queue it up. Outside contributions are always welcome, of course :)

PR #87, opened 2020-06-07
pass only the necessary bits to ref_stats() for clarity

ozancaglayan:
If it is okay to break internal API, this makes it clear that the only information required from the output side for ref_stats is the length.

TODO items:

- [ ] Add multi-ref support to chrF
- [ ] Check whether smoothing method and values are taken into account correctly for BLEU
- [ ] Fix the handling of `num_refs` across the codebase + metrics + signature


ozancaglayan:
Let's not merge this now as I am doing the tokenizer work under this branch.

ozancaglayan:
Will potentially close #5 #27 and #85

I ran `test.sh` with MECAB enabled and all 154 tests passed. Although I tried to be attentive as much as possible, I don't know if `intl` and `zh` tokenizers are tested with the test suite.

PR #93, opened 2020-06-25
Added support for character-based tokenisation.

cfedermann:
Hi, @mjpost,

I had added this `char-based` tokenisation option for OCELoT recently. It gives a cheap way to compute more useful scores for non-word-broken language scripts and can be used as a fallback when there is no tokenisation available (yet).

Cheers and best, have a good day,
   Christian


ozancaglayan:
Could you rebase this on the new master by adding a new Tokenizer class?

mjpost:
I would also suggest we just call it "char" (i.e., `-tok char` instead of `-tok char-based`).

mjpost:
Closed with 32f433f092a880e031951d5bd42558bcde3736af.

Issue #86, opened 2020-06-04
Crashes when run without installation


emjotde:
Hi,
Used to be possible to run `sacrebleu` directly from the cloned folder like:

```
git clone https://github.com/mjpost/sacrebleu
./sacrebleu/sacrebleu/sacrebleu.py -h
```

This dies now with:
```
Traceback (most recent call last):
  File "./sacrebleu/sacrebleu/sacrebleu.py", line 41, in <module>
    from .tokenizer import TOKENIZERS, TokenizeMeCab
SystemError: Parent module '' not loaded, cannot perform relative import
```

mjpost:
Does `python -m sacrebleu` work (from the sacrebleu root dir)?

emjotde:
Yep, that works.

mjpost:
It was already there :) In your defense, I did break backward compatibility.

emjotde:
Ah, but I meant it could be run from anywhere. Sorry, that was unclear. So any relative path would work before.

mjpost:
So, I think you are asking for a CLI script that you don't need to be in the directory to run?

emjotde:
yes, that worked like here:
https://github.com/marian-nmt/marian-examples/blob/c19b7814d71febf1053bd93af6ac314b46204092/training-basics-sentencepiece/run-me.sh

If you don't want to support that then OK. I just assumed it's a regression.

cfedermann:
Instead of running:

    ./sacrebleu/sacrebleu/sacrebleu.py -h

run it like this:

    python -m sacrebleu.sacrebleu ./sacrebleu/sacrebleu/sacrebleu.py -h

If you know the absolute path, use `PYTHONPATH`:

    PYTHONPATH=/path/to/sacrebleu python -m sacrebleu.sacrebleu /path/to/sacrebleu/sacrebleu/sacrebleu.py -h

emjotde:
I am sure we can come up with a version that mentions the module path more than just three times :)

mjpost:
It’s like the opposite of the Obfuscated Perl contests—call it Prolix Python. 

I am not sure if Python allows a “sacrebleu.py” in the root dir but let me see what I can do.

emjotde:
Thanks! Christian's solution reminds me of the darks times when I last had to use java and CLASSPATH some 20(?) years ago.

ozancaglayan:
[...]

PR #95, opened 2020-07-03
Allows calling the script as a standalone utility (#86)

ozancaglayan:
I tested it with installed sacrebleu, with `-m sacrebleu` from root dir, calling the script from arbitrary places (when installed and not-installed). It seems to work.

### DIFF
--- a/sacrebleu/sacrebleu.py
+++ b/sacrebleu/sacrebleu.py
@@ -25,7 +25,15 @@ See the [README.md] file for more information.
 import io
 import sys
 import logging
+import pathlib
 import argparse
+
+# Allows calling the script as a standalone utility
+# See: https://github.com/mjpost/sacrebleu/issues/86
+if __package__ is None and __name__ == '__main__':
+    parent = pathlib.Path(__file__).absolute().parents[1]
+    sys.path.insert(0, str(parent))
+    __package__ = 'sacrebleu'
 
 from .tokenizers import TOKENIZERS, DEFAULT_TOKENIZER
 from .dataset import DATASETS, DOMAINS, COUNTRIES, SUBSETS


martinpopel:
I tested this by uninstalling sacrebleu, then `git clone https://github.com/mjpost/sacrebleu; git pull https://github.com/ozancaglayan/sacrebleu.git fix-86` and then executed `python3 -m sacrebleu -V`.
First, it failed with `No module named 'MeCab'`, but this is expected that I need to install the dependencies first.
After installing MeCab, it works (prints `__main__.py 1.4.12`), but only from the directory with `sacrebleu.py`.
If `sacrebleu.py` is e.g. in `path/to/sacrebleu`, I need to run it as `python3 -m path.to.sacrebleu.sacrebleu -V`.

I thought the goal was to run it as `path/to/sacrebleu/sacrebleu.py -V`.
Unfortunately, this does not work for me -- it fails with
```
Traceback (most recent call last):
  File "path/to/sacrebleu/sacrebleu.py", line 38, in <module>
    from .tokenizers import TOKENIZERS, DEFAULT_TOKENIZER
SystemError: Parent module 'sacrebleu' not loaded, cannot perform relative import
```

ozancaglayan:
Interesting, it was working for me when I sent this but I will recheck what's going on. Thanks!

ozancaglayan:
This still works for me (with and without installing the package)

```bash
$ git clone git clone https://github.com/mjpost/sacrebleu
$ cd sacrebleu
$ gh pr checkout 95
# Is it installed?
$ which sacrebleu
# Check python path
$ python -c 'import sys; print("\n".join(sys.path))'

/home/ozan/anaconda3/lib/python37.zip
/home/ozan/anaconda3/lib/python3.7
/home/ozan/anaconda3/lib/python3.7/lib-dynload
/home/ozan/anaconda3/lib/python3.7/site-packages
/home/ozan/git/transformers

# OK, go back to home
$ cd

$ ./sacrebleu/sacrebleu/sacrebleu.py -V
sacrebleu.py 1.4.12

$ cd sacrebleu
$ python -m sacrebleu -V
# this should be fixed BTW, it should return a static string for correctness
__main__.py 1.4.12

martinpopel:
When replicating the steps on another computer it worked, but on my notebook it still fails. Maybe there is something wrong with my python installation, but unfortunately, I am too busy to delve into it now. You can merge this and we will see if other users can use it as a standalone script or if the file a new issue.

ozancaglayan:
Re-added this into 1.4.14 PR since it's a trivial commit. Thanks!

Issue #98, opened 2020-07-15
Difference score for one sentence with sentence_blue() vs Command Line --sentence-level


ymoslem:
**SacreBLEU version:** 1.4.12
**Python:** 3.7

**Issue:** Using Python `sentence_blue()` and the Command Line with the `--sentence-level` flag (or `-sl`) for only one sentence gives different scores.
**Question:** I need `sentence_blue()` from Python to give the same output as the Command Line with the `--sentence-level` flag (or `-sl`), what should I do?

## From the Command Line:

```
echo "Producția de zahăr primă va fi exprimată în ceea ce privește zahărul alb;" | sacrebleu -sl <(echo "producţia de zahăr brut se exprimă în zahăr alb;")
```
**Output:**
```
BLEU+case.mixed+numrefs.1+smooth.exp+tok.13a+version.1.4.12 = 8.5 35.7/15.4/4.2/2.3 (BP = 1.000 ratio = 1.400 hyp_len = 14 ref_len = 10)
```

## From Python
**Using `sentence_blue()`**
```
ref = "producţia de zahăr brut se exprimă în zahăr alb;"
sys = "Producția de zahăr primă va fi exprimată în ceea ce privește zahărul alb;"

bleu = sacrebleu.sentence_bleu(sys, ref)
print(bleu.score)
```
**Output:**
```
0.0
```
**Using `corpus_blue()`**
```
ref = [["producţia de zahăr brut se exprimă în zahăr alb;"]]
sys = ["Producția de zahăr primă va fi exprimată în ceea ce privește zahărul alb;"]

bleu = sacrebleu.corpus_bleu(sys, ref)
print(bleu.score)
```
**Output:**
```
8.493098745313148
```

martinpopel:
This is related to #64, but slightly more complicated.

First, note the [disclaimer](https://github.com/mjpost/sacrebleu/blob/e22640/sacrebleu/compat.py#L66-L67): *computing BLEU on the sentence level is not its intended use, BLEU is a corpus-level metric.*
If you need a sentence-level metric, try e.g. chrF (`sacrebleu -m chrf`).

Second, you should use `sacrebleu.sentence_bleu(sys, [ref])` instead of `sacrebleu.sentence_bleu(sys, ref)` because it takes a list of reference translations for a given sentence. (I fixed the documentation in #99.)

[...]

PR #99, opened 2020-07-16
sentence_score needs a list of references, not a single string

martinpopel:
If a single string is provided it is split into characters and each character is treated as one reference, which may be very confusing and difficult to debug.
The type hint `List[str]` could help if everyone is using some type checking such as `mypy`.

### DIFF
--- a/sacrebleu/compat.py
+++ b/sacrebleu/compat.py
@@ -67,7 +67,7 @@ def sentence_bleu(hypothesis: str,
     BLEU is a corpus-level metric.
 
     :param hypothesis: Hypothesis string.
-    :param reference: Reference string.
+    :param references: List of reference strings.
     :param smooth_method: The smoothing method to use
     :param smooth_value: For 'floor' smoothing, the floor value to use.
     :param use_effective_order: Account for references that are shorter than the largest n-gram.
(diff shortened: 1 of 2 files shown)


ozancaglayan:
Either we bail out and assert/throw exception or we handle the inputs accordingly and convert single strings to list of string etc. I think this is done in some BLEU functions through `isinstance` calls but I couldn't grasp which cases do they handle exactly. This is quite a delicate issue which should be fixed in a nice/clarified way I guess.

martinpopel:
The main problem this PR solves is that the documentation was wrong.
In addition, I added the assertion to prevent a bug (which happened in #98), which was quite tricky because the string was split into characters, each character was treated as one reference and a non-zero (but wrong, of course) BLEU score was returned.

> convert single strings to list of string etc. I think this is done in some BLEU functions through isinstance calls 

Yes, this is currently done in `bleu.py:corpus_score()`, but not in `bleu.py:sentence_score()`.
An alternative solution would be to accept a single reference string even here, but once we allow this in the API, we will have to keep it forever. I would prefer non-bloated API with no "robustness to user errors". I mean I would prefer to move the `isinstance` hacks to `compat.py` before someone starts using the new API with a single reference instead of a list of references - but that can be done in a different PR after a discussion. This PR was meant as a non-controversial bugfix.

Issue #15, opened 2017-11-22
SGML entities de-escaping in tokenization


martinpopel:
https://github.com/mjpost/sacreBLEU/blob/b38690e1537cd4719c3517ef77c8255c5a107cc8/sacrebleu.py#L396-L399

First, there is a bug: all four entities `&quot; &amp; &lt; &gt;` are converted to double quotes. Probably a copy-paste error ([not present](https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v14.pl#L934-L937) in the original Perl implementation).

Second, I think we should delete this completely. The de-escaping was needed in the original implementation because the translation and reference files were in SGML (*.sgm) format. SacréBLEU expects plain-text input (or API calls), so this is not needed. I think it is a responsibility of a modern MT system to clean the data (ideally the training data) and produce human-readable sentences (i.e. without escaped html/sgml entities).

Similarly, if the input format is expected to be one sentence per line, there is no need for `replace('-\n', '')`, but this does not matter if there are no newlines in the string, it just obfuscates the code.

And guess what is my opinion on `replace('<skipped>', '')`.

mjpost:
Nice catch. I'm going to fix this instead of remove it, though, because my goal with the tokenization_13a is to replicate mteval_13a.pl, which is used by Moses. This bug has no effect on the WMT datasets, which hasn't used these codes since 2009. There are, however, other test sets that I may add someday that do use these (for example, the Chinese NIST datasets from the mid-2000s).

martinpopel:
If there are *.sgm test sets which use &quot; etc. then these should be de-escaped already when exporting the sgml to plain-text.

ozancaglayan:
this seems to be already handled, right?

martinpopel:
The obvious bug was fixed (converting all entities to double quotes), but the de-escaping of SGML entities (converting `&amp;` to &, `&lt;` to < etc.) is still present in the default tokenizer.
What I suggested in this issue is that the entities should be de-escaped when reading SGML/XML references, but it should not be part of the tokenizer. This is still not done, which is why the issue is still open.

ozancaglayan:
[...]

PR #102, opened 2020-07-17
Add TER

ales-t:
Hi @mjpost , please check out this work-in-progress version of TER in sacrebleu. Better late than never I guess :slightly_smiling_face: 

I think the core is basically there and it works but I'm not sure about some issues.

For instance, which parameters to expose in the CLI, and how to handle tokenization (TER has its own tokenizer which I re-implemented but we should probably allow users to switch to a different one -- for now I simply always use the TER tokenizer).

Also, I haven't given much thought to the signatures.

I think tests cover the metric pretty well. I'm planning to add tests of the TER tokenizer next week.

I'll also need to discuss with Memsource colleagues what license header to add to the new files.

EDIT: To justify the complex code: existing implementations of TER in Python, while simpler (such as `pyter`), do not produce results consistent with Tercom and can be very slow for some inputs. There are many small design decisions in the original Tercom implementation, some of them are arbitrary, and these have to be taken into account. Also, some additional optimization was required to make the code reasonably fast.

### DIFF
--- a/sacrebleu/compat.py
+++ b/sacrebleu/compat.py
@@ -3,8 +3,8 @@
 from typing import Union, Iterable, List
 from argparse import Namespace
 
-from .tokenizers import DEFAULT_TOKENIZER
-from .metrics import BLEU, CHRF, BLEUScore, CHRFScore
+from .tokenizers import DEFAULT_TOKENIZER, TOKENIZERS
+from .metrics import BLEU, CHRF, TER, BLEUScore, CHRFScore, TERScore
 
 
 ######################################################################
@@ -121,4 +121,52 @@ def sentence_chrf(hypothesis: str,
     args = Namespace(
         chrf_order=order, chrf_beta=beta, chrf_whitespace=not remove_whitespace, short=False)
     metric = CHRF(args)
+    return metric.sentence_score(hypothesis, references)
+
+
+def corpus_ter(hypotheses: Iterable[str],
+               references: List[Iterable[str]],
+               normalized: bool = False,
+               no_punct: bool = False,
+               asian_support: bool = False,
+               case_sensitive: bool = False) -> TERScore:
+    """
+    Computes TER on a corpus.
+
+    :param hypotheses: Stream of hypotheses.
+    :param references: Stream of references.
+    :param normalized: Enable character normalization.
+    :param no_punct: Remove punctuation.
+    :param asian_support: Enable special treatment of Asian characters.
+    :param case_sensitive: Enable case sensitivity.
+    :return: A `TERScore` object.
+    """
+    tokenizer = TOKENIZERS['tercom'](
+        normalized=normalized, no_punct=no_punct,
+        asian_support=asian_support, case_sensitive=case_sensitive)
+    metric = TER(Namespace(), tokenizer)
+    return metric.corpus_score(hypotheses, references)
+
+
+def sentence_ter(hypothesis: str,
+                 references: Iterable[str],
+                 normalized: bool = False,
+                 no_punct: bool = False,
+                 asian_support: bool = False,
+                 case_sensitive: bool = False) -> TERScore:
+    """
+    Computes TER on a single sentence pair.
+
+    :param hypothesis: Hypothesis string.
+    :param references: Reference string(s).
+    :param normalized: Enable character normalization.
+    :param no_punct: Remove punctuation.
+    :param asian_support: Enable special treatment of Asian characters.
+    :param case_sensitive: Enable case sensitivity.
+    :return: A `TERScore` object.
+    """
+    tokenizer = TOKENIZERS['tercom'](
+        normalized=normalized, no_punct=no_punct,
+        asian_support=asian_support, case_sensitive=case_sensitive)
+    metric = TER(Namespace(), tokenizer)
     return metric.sentence_score(hypothesis, references)
--- a/sacrebleu/compat.py
+++ b/sacrebleu/compat.py
@@ -141,10 +141,10 @@ def corpus_ter(hypotheses: Iterable[str],
     :param case_sensitive: Enable case sensitivity.
     :return: A `TERScore` object.
     """
-    tokenizer = TOKENIZERS['tercom'](
+    args = Namespace(
         normalized=normalized, no_punct=no_punct,
         asian_support=asian_support, case_sensitive=case_sensitive)
-    metric = TER(Namespace(), tokenizer)
+    metric = TER(args)
     return metric.corpus_score(hypotheses, references)
 
 
@@ -165,8 +165,8 @@ def sentence_ter(hypothesis: str,
     :param case_sensitive: Enable case sensitivity.
     :return: A `TERScore` object.
     """
-    tokenizer = TOKENIZERS['tercom'](
+    args = Namespace(
         normalized=normalized, no_punct=no_punct,
         asian_support=asian_support, case_sensitive=case_sensitive)
-    metric = TER(Namespace(), tokenizer)
+    metric = TER(args)
     return metric.sentence_score(hypothesis, references)
--- a/sacrebleu/metrics/ter.py
+++ b/sacrebleu/metrics/ter.py
@@ -68,6 +68,11 @@ class TERScore(BaseScore):
         self.prefix = 'TER'
 
     def format(self, width=2, score_only=False, signature=''):
+        # the default width of 1 is too small for TER, it's reported with 3
+        # decimal places in matrix.statmt.org
+        if width == 1:
+            width = 3
+
         if score_only:
             return '{0:.{1}f}'.format(self.score, width)
 
--- a/sacrebleu/metrics/ter.py
+++ b/sacrebleu/metrics/ter.py
@@ -1,0 +1,565 @@
+import math
+from typing import List, Tuple, Dict, Union, Iterable
+from itertools import zip_longest
+
+from ..tokenizers import TOKENIZERS
+from ..utils import my_log
+from .base import BaseScore, Signature
+
+# Translation edit rate (TER).
+#
+# Based partly on the PyTer package by Hiroyuki Tanaka (MIT license).
+#
+# https://github.com/aflc/pyter
+#
+# Both the algorithm and the preprocessing try to mimic the original TER Java
+# implementation (Tercom):
+#
+# https://github.com/jhclark/tercom
+
+_COST_INS = 1
+_COST_DEL = 1
+_COST_SUB = 1
+
+# Tercom-inspired limits
+_MAX_SHIFT_SIZE = 10
+_MAX_SHIFT_DIST = 50
+_BEAM_WIDTH = 20
+
+# Our own limits
+_MAX_CACHE_SIZE = 10000
+_MAX_SHIFT_CANDIDATES = 1000
+
+_OP_INS = 'i'
+_OP_DEL = 'd'
+_OP_NOP = ' '
+_OP_SUB = 's'
+_OP_UNDEF = 'x'
+
+
+class TERScore(BaseScore):
+    def __init__(self, num_edits, ref_length):
+        score = num_edits / ref_length
+        super().__init__(score)
+
+        self.num_edits = num_edits
+        self.ref_length = ref_length
+
+    def format(self, width=2, score_only=False, signature=''):
+        return str(self.score)
+
+    def __gt__(self, other):
+        if type(other) != TERScore:
+            raise TypeError()
+
+
+class TERSignature(Signature):
+    def __init__(self, args):
+        super().__init__(args)
+
+
+class TER:
+    def __init__(self, args):
+        self.tokenizer = TOKENIZERS[args.tokenize]()
+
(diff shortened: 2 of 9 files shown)


sacrebleu/tokenizers/tokenizer_ter.py
```diff
@@ -0,0 +1,148 @@
...
+    # language-independent (general) part
+
+    # strip end-of-line hyphenation and join lines
+    sent = re.sub(r"\n-", "", sent)
+
+    # join lines
+    sent = re.sub(r"\n", " ", sent)
+
+    # handle XML escaped symbols
+    sent = re.sub(r"&quot;", "\"", sent)
```
martinpopel, line 16: Ideally, this XML entities de-escaping should not be in the tokenizers, but in reading references (https://github.com/mjpost/sacrebleu/blob/e22640c996ca48d018fd974ba26a052180f4c77d/sacrebleu/utils.py#L53) (system outputs are provided already as plaintext to SacreBLEU, so the de-escaping should *not* be applied on these).
However, the same bug (#15) is in BLEU's default tokenizer [13a](https://github.com/mjpost/sacrebleu/blob/master/sacrebleu/tokenizers/tokenizer_13a.py), so I suggest to merge this PR as is and fix this later.

ales-t:
I have added test of the tokenizer and I've done some minor changes. I think the PR is more or less ready, I just wonder what to do about:

* configuration of TER tokenization (and its CLI parameters)
* signatures

Can you please give me some hints?

I guess we could also merge this as-is and worry about improvements later, it's up to you.

sacrebleu/tokenizers/tokenizer_ter.py
ales-t: Thanks for the suggestion. Would it make sense to simply remove these lines from the tokenizer then?

Note though that by default, this normalization is disabled, so it won't be applied unless the user configures it (which can't be done through command-line options for the time being anyway).

Also, if someone was looking (for some obscure reason) for a Python tokenizer that's consistent with Tercom, they could simply use the class `TercomTokenizer` and get the desired result. If we removed this normalization, the tokenizer would no longer be exactly the same.

ozancaglayan:
Hi,

I'll go over this shortly but I have one particular concern about the way tokenizers are used in sacreBLEU. BLEU allows choosing the tokenizer so far amongst several options (which may be arguably bad for reproducibility as well). For TER, this shouldn't be configurable to replicate the original TERCOM right? In that case, we should maybe not expose tercom tokenizer and check that the signature of the tokenizer/metric and the tokenizer itself is always the same regardless of the requested metric from cmdline.

bricksdont:
Hi, this is a great PR!

Would it perhaps make sense to also expose TER corpus and sentence-level functions in `compat.py`? Here:

https://github.com/mjpost/sacrebleu/blob/master/sacrebleu/compat.py

That would make TER more accessible from Python

ales-t:
> Hi, this is a great PR!
> 
> Would it perhaps make sense to also expose TER corpus and sentence-level functions in `compat.py`? Here:
> 
> https://github.com/mjpost/sacrebleu/blob/master/sacrebleu/compat.py
> 
> That would make TER more accessible from Python. If I understand correctly, without such a function one would have to create an instance of this metric first and then call its `score` functions.

Hi @bricksdont , I've added the `compat` functions and briefly tested them, please check the updated MR.

ales-t:
> Hi,
> 
> I'll go over this shortly but I have one particular concern about the way tokenizers are used in sacreBLEU. BLEU allows choosing the tokenizer so far amongst several options (which may be arguably bad for reproducibility as well). For TER, this shouldn't be configurable to replicate the original TERCOM right? In that case, we should maybe not expose tercom tokenizer and check that the signature of the tokenizer/metric and the tokenizer itself is always the same regardless of the requested metric from cmdline.

Hi @ozancaglayan , `TER` at this point ignores the command-line setting of the tokenizer option, and always instantiates its own `TercomTokenizer`. So users can't force different tokenization for TER. This is IMO not intuitive (why is my setting of `--tokenize` ignored?) but I guess it's a sensible solution. Maybe there should be a warning when TER is used along with `--tokenize`?

On the other hand, users are allowed to use `TercomTokenizer` with BLEU. This perhaps doesn't make as much sense but as you say, users can choose custom tokenizers for BLEU already, so why not have this option too.

I changed the signature of TER to include the tokenizer signature. Are there other changes that you would propose?

ozancaglayan:
Hi,

One invasive solution which may turn out to be better in the long run is to use `argparse`'s sub-commands to segregate TER, CHRF and BLEU into different subcommands i.e.

```
sacrebleu bleu --tokenize v13a <etc.>
sacrebleu ter --some_ter_parameters 1.0 <etc.>
sacrebleu chrf --alpha 0.6 <etc.>
```

This will (1) avoid to prefix metric-specific arguments in argparse with things like `--chrf-foo=bar`, (2) make it clear that `--tokenize` only applies to BLEU and not to TER, (3) some other potential benefits as well in isolating metrics in a neat way such as when we implement significance testing per metric, or to improve readability in general.

But this probably requires more discussion and comments from people (@mjpost @martinpopel). Otherwise, ignoring --tokenize as you said seemed okay to me. I'll go over the latest changes as well, thanks!

ales-t:
Hi, is there any progress with the review? I'm happy to make more changes if needed but I'm not sure there are any outstanding issues to address at this point.

martinpopel:
I am for merging this now (and I will do so tomorrow, if there are no objections).

@ozancaglayan I am not sure about the idea with subcommands. I understand its advantages, but there are also disadvantages:
* Either we allow omitting `bleu` as the default subcommand (but then it may not be so obvious that `tokenize` only applies to BLEU, which is another issue I am not sure about) or we will break backward compatibility.
* It won't be possible anymore to evaluate multiple metrics within one sacrebleu execution.

Anyway, I think the subcommands should be discussed in a separate issue/PR, after this PR is merged.

mjpost:
Sounds good, @martinpopel. I'll review today.

ozancaglayan:
I'll do a pass now as well

ales-t:
I discovered some minor issues with the code today. Please don't merge the PR yet, I will try to fix them tomorrow and I'll let you know. Sorry for the complications.

sacrebleu/compat.py
```diff
@@ -122,3 +122,51 @@ def sentence_chrf(hypothesis: str,
...
+
+    :param hypotheses: Stream of hypotheses.
+    :param references: Stream of references.
+    :param normalized: Enable character normalization.
+    :param no_punct: Remove punctuation.
+    :param asian_support: Enable special treatment of Asian characters.
+    :param case_sensitive: Enable case sensitivity.
+    :return: A `TERScore` object.
+    """
+    tokenizer = TOKENIZERS['tercom'](
```
ozancaglayan, line 144: Is there a specific reason why you wouldn't pass these to the metric and let it create the tokenizer internally?

sacrebleu/tokenizers/tokenizer_ter.py
```diff
@@ -0,0 +1,170 @@
...
+            sent = re.sub(r"^\s+", "", sent)  # no leading space
+            sent = re.sub(r"\s+$", "", sent)  # no trailing space
+
+        if self._no_punct:
+            sent = _remove_punct(sent)
+            if self._asian_support:
+                sent = _remove_asian_punct(sent)
+
+        return sent
+
```
ozancaglayan, line 162: there's a trailing whitespace/tab here at line 162.

sacrebleu/metrics/ter.py
```diff
@@ -0,0 +1,592 @@
...
+                if best_num_edits is None or num_edits < best_num_edits:
+                    best_num_edits = num_edits
+
+            total_edits += best_num_edits
+            sum_ref_lengths += (ref_lengths / len(refs))
+
+        return TERScore(total_edits, sum_ref_lengths)
+
+    def sentence_score(self, hypothesis: str, references: List[str]) \
+            -> TERScore:
```
ozancaglayan, line 129: We (and you as well :)) already have longer lines than this so I would lean towards not using the continuation marker and move `-> TERScore` above. (In remaining codes as well).

mjpost:
@ales-t sounds good. I'll wait to hear from you before doing my review.

sacrebleu/compat.py
ales-t: Just to be sure, you're suggesting that the metric should accept these arguments (`normalized`, `no_punct` etc.), correct?

The constructor currently accepts `args` (which it ignores) and an optional tokenizer. Which would be preferable?

* Have `TER` accept `args` and a bunch of additional options with pre-set defaults.
* Add new CLI options and include these settings in `args` directly.

I feel that the second option, while cleaner, would lead to cascade of changes: TER tokenizer can be used with other metrics as well, so perhaps it should also be configurable from the CLI. So maybe the options should be called `--ter-tokenizer-normalize`. But then, all tokenizer classes would need to accept `args` as well and wherever they're instantiated, we would need to provide the configuration...

So maybe option 1?

sacrebleu/tokenizers/tokenizer_ter.py
ales-t: Removed.

sacrebleu/metrics/ter.py
ales-t: Yes, good point. Most of the code was taken from an internal code base where we try to follow the Google style guide. I removed the backslashes, hopefully I didn't miss any.

ales-t: I also converted the original Google-style docstrings into RST.

sacrebleu/compat.py
ozancaglayan: BLEU and CHRF metrics receive their own arguments through `args` and then parse them. CHRF does not have a tokenizer but BLEU has one and if you look to `BLEU.__init__()`, that's where it is initialized. It just didn't make sense to me that BLEU creates it internally whereas TER creates the tokenizer externally and passes nothing to TER as you mentioned.

I would go for option 1: Let TER consume the arguments through `args` as is done in BLEU and CHRF. The defaults can be stored in class variables as in `BLEU`. The `TER` signature as of now adds an explicit `tokenizer` argument as well, which was not done for BLEU. At some point, we'll make all these arguments explicit with default values in `__init__` also. The encapsulating through `args` was a dirty path that I choose.

Do we really want to make TER tokenizer reusable by other metrics through CLI? It is still accessible through the programming interface if one really needs it? What do you think?

ales-t: Cool, so the idea is to encapsulate the settings in `args` which (when called from `compat`) would contain keys that are not actual CLI options, correct? And when TER is instantiated from the main script, `args` would not contain these so we would fall back to defaults?

I'll remove TER from the available tokenizers, I think that makes sense.

As for the signature, if we want to only pass `args` to it, then it would also need to unpack them and figure out whether defaults were used or whether custom settings were provided (through `compat`). I think that makes it a bit less clean but I guess it's a good workaround for the time being.

ales-t: You can check the updated solution: https://github.com/mjpost/sacrebleu/pull/102/commits/2f1705f70618ec780a8298b76318d3e4981ad426

ales-t:
@mjpost I fixed both issues.

* The first problem was a tiny difference in score for one of the tested submissions (the difference was 1 edit operation on the whole test set). Slightly increasing the default beam width fixed the problem (this implementation is slightly different than Tercom beam search but it really only shows on quite extreme inputs, as was the case here).
* The default formatting width was too small for TER which is typically reported with 3 decimal places. TER now overrides the default setting in a similar way as `chrf3`.

At this point, I'd say it's ready for review.

mjpost:
Also, @ales-t, would you mind merging in master so this is up to date? We'll merge this within the next day or two, so don't worry that your work will be in vain.

ales-t:
Hi @mjpost , thanks! I've updated the branch so it should be ready for merging.

ozancaglayan (approved):

ozancaglayan:
Hi @ales-t , Is the default implementation of TER uncased as well? I was going on the code to potentially unify the lowercase arguments of all three metrics but noticed that TER currently lowercases sentences by default

ales-t:
Hi @ozancaglayan , I find it a bit weird as well.

Frankly, the default settings of TER are unfortunate in various ways (for instance, it has support for Asian languages but that's disabled by default, so matrix.statmt.org will happily report TER higher than 1 for en-zh, such as 28 as in this case: http://matrix.statmt.org/systems/show/4472).

I would kind of lean towards keeping the tercom defaults for the sake of consistency with WMT. All the settings are available so informed users can set the metric as they like but they will then probably be aware that this is not the default, and report it accordingly in their papers.

ozancaglayan:
Okay thanks. Your suggestion is reasonable but those options are only available through the API and not cmdline as of now. Maybe we can add them as `--ter-option value` then.

ozancaglayan:
Okay thanks. Your suggestion is reasonable but those options are only available through the API and not cmdline as of now. Maybe we can add them as `--ter-option value` then.

PR #106, opened 2020-07-24
Fix wrong return type of corpus_bleu

heiwais25:
I think the return type of `corpus_bleu` is wrong. It should be `BLEUScore` as it is shown on the docstring.

### DIFF
--- a/sacrebleu/compat.py
+++ b/sacrebleu/compat.py
@@ -17,7 +17,7 @@ def corpus_bleu(sys_stream: Union[str, Iterable[str]],
                 force=False,
                 lowercase=False,
                 tokenize=DEFAULT_TOKENIZER,
-                use_effective_order=False) -> BLEU:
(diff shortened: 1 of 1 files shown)


ozancaglayan:
Fairly trivial fix for the signature, we can merge after an approval.

mjpost:
@ozancaglayan, you should be authorized to approve. I don't think we'll do a new release yet, but you could approve, merge, and then add a stub to CHANGELOG.

ozancaglayan:
Ok I've merged this, thanks! @mjpost Since master is protected, I should pass through branch-PR-merge even for modifying the Changelog, right? We can write Changelog at once then before release by looking back into the commit history.

martinpopel:
@ozancaglayan I think you don't need to start a new PR - you could add a commit to this PR and than squash and merge. We can try it next time.

PR #107, opened 2020-07-25
Fix Japanese handling

polm:
This changes the Japanese tokenizer to use versions of mecab-python3 1.0
or greater. This means the package will work on Windows.

However, since the Japanese tokenizer pulls in heavy dependencies and
isn't necessary unless you're dealing with Japanese, I moved it to
optional dependencies. You can install sacrebleu with Japanese support
like below:

    pip install sacrebleu[ja]

That will install mecab-python3 and ipadic.

This also includes basic tests to check that the tokenization is as
exepcted for IPAdic.

You may need to make changes to your automated testing setup to install the Japanese dependencies.

### DIFF
--- a/sacrebleu/tokenizers/tokenizer_ja_mecab.py
+++ b/sacrebleu/tokenizers/tokenizer_ja_mecab.py
@@ -22,8 +22,6 @@ class TokenizerJaMecab(NoneTokenizer):
         self.tagger = MeCab.Tagger(ipadic.MECAB_ARGS + " -Owakati")
 
         # make sure the dictionary is IPA
-        # sacreBLEU is only compatible with 0.996.5 for now
-        # Please see: https://github.com/mjpost/sacrebleu/issues/94
         d = self.tagger.dictionary_info()
         assert d.size == 392126, \
             "Please make sure to use the IPA dictionary for MeCab"
--- a/sacrebleu/tokenizers/tokenizer_ja_mecab.py
+++ b/sacrebleu/tokenizers/tokenizer_ja_mecab.py
@@ -1,20 +1,33 @@
 # -*- coding: utf-8 -*-
-
-import MeCab
+try:
+    import MeCab
+    import ipadic
+except ModuleNotFoundError:
+    # Don't fail until the tokenizer is actually used
+    MeCab = None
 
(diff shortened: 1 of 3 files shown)


polm:
I made the necessary changes to Travis and it looks like the tests are passing now.

sacrebleu/tokenizers/tokenizer_ja_mecab.py
```diff
@@ -1,20 +1,33 @@
...
 
 class TokenizerJaMecab(NoneTokenizer):
     def __init__(self):
-        self.tagger = MeCab.Tagger("-Owakati")
+        if MeCab is None:
+            raise RuntimeError(FAIL_MESSAGE)
+        self.tagger = MeCab.Tagger(ipadic.MECAB_ARGS + " -Owakati")
 
         # make sure the dictionary is IPA
         # sacreBLEU is only compatible with 0.996.5 for now
```
ozancaglayan, line 25: Can you remove these comments related to 0.996.5 then?

ozancaglayan: Also, can you rebase your branch as master moved forward by 1 commit in the meanwhile?

martinpopel: @ozancaglayan I think you can merge master into this polm:japanese-fix branch yourself, e.g. by clicking on the "Update branch" button on GitHub. @mjpost prefers to squash all commits in the PR, so in the end it does not matter if you rebase the PR or merge in master.

polm: I removed the comments and updated the branch.

polm:
> So pip install sacrebleu[ja] does a full install with mecab, right?

That's correct.

ozancaglayan:
Thanks for the PR. I will test this tomorrow a bit further and then @mjpost could do a release to fix Windows installation issue.

ozancaglayan:
@polm why does `tagger.version()` still returns 0.996? Shouldn't this reflect MeCab' version?

polm:
That's the version of the underlying C++ library, which is not the same as the version number used for `mecab-python3`, the Python wrapper.

PR #109, opened 2020-07-30
Added WMT20 newstest

mjpost:


sacrebleu/dataset.py
```diff
@@ -22,6 +22,38 @@
...
+        'fr-de': ['sgm/newstest2020-frde-src.fr.sgm', 'sgm/newstest2020-frde-ref.de.sgm'],
+        'iu-en': ['sgm/newstest2020-iuen-src.iu.sgm', 'sgm/newstest2020-iuen-ref.en.sgm'],
+        'ja-en': ['sgm/newstest2020-jaen-src.ja.sgm', 'sgm/newstest2020-jaen-ref.en.sgm'],
+        'km-en': ['sgm/newstest2020-kmen-src.km.sgm', 'sgm/newstest2020-kmen-ref.en.sgm'],
+        'pl-en': ['sgm/newstest2020-plen-src.pl.sgm', 'sgm/newstest2020-plen-ref.en.sgm'],
+        'ps-en': ['sgm/newstest2020-psen-src.ps.sgm', 'sgm/newstest2020-psen-ref.en.sgm'],
+        'ru-en': ['sgm/newstest2020-ruen-src.ru.sgm', 'sgm/newstest2020-ruen-ref.en.sgm'],
+        'ta-en': ['sgm/newstest2020-taen-src.ta.sgm', 'sgm/newstest2020-taen-ref.en.sgm'],
+        'zh-en': ['sgm/newstest2020-zhen-src.zh.sgm', 'sgm/newstest2020-zhen-ref.en.sgm'],
+        'de-en': ['sgm/newstestB2020-deen-src.de.sgm', 'sgm/newstestB2020-deen-ref.en.sgm'],
```
ozancaglayan, line 51: Curiosity: How does newstestB2020 gets selected for 'de-en'? I would have expected that to be another test set although I don't know what that B stands for =)

ozancaglayan:
I'll merge my other PR into this

sacrebleu/dataset.py
mjpost: Oh! I didn't notice this. We'll have to choose one of them. Let me check.

ozancaglayan:
Can you also add my to test.pypi.org sacreBLEU?

sacrebleu/dataset.py
mjpost: So, for some datasets, there are two references. I just sent an email to find out whether (a) these two references should be used by default or (b) they should be used as a separate dataset.

mjpost:
Done. It looks like I haven't used that site in quite some time (release 1.0.2).

ozancaglayan:
I already tested v1.4.13 through testpypi without the last commit, with and without `[ja]` as well. It seems to work, the tests pass :crossed_fingers: 

If this branch is OK to merge, then I can push to normal pypi as well.

mjpost:
Can you tell me what `[ja]` does? I have never seen this before and it looks to me like a shell wildcard.

ozancaglayan:
This is an `extra` suffix, which allows a layered installation experience. @polm defined extra deps for `[ja]` in `setup.py`.
See: https://stackoverflow.com/questions/52474931/what-is-extra-in-pypi-dependency

ozancaglayan:
Should I proceed?

mjpost:
I haven't heard back on whether WMT used the two references, but I think it's okay to proceed the way I have it.

ozancaglayan:
OK, I pushed v1.4.13 to pypi. Hopefully, I did not mess things up this time :crossed_fingers:

mjpost:
Works for me, thanks, @ozancaglayan!

PR #110, opened 2020-07-31
Use ImportError instead of ModuleNotFoundError to support python3.5

thomasZen:
Sacrebleu does not support python 3.5 anymore, because it uses [ModuleNotFoundError](https://docs.python.org/3/library/exceptions.html?highlight=modulenotfounderror#ModuleNotFoundError). I replaced it with [ImportError](https://docs.python.org/3/library/exceptions.html?highlight=modulenotfounderror#ModuleNotFoundError) which is present since python 3.3. (ModuleNotFoundError is a subclass of ImportError).

### DIFF
--- a/sacrebleu/tokenizers/tokenizer_ja_mecab.py
+++ b/sacrebleu/tokenizers/tokenizer_ja_mecab.py
@@ -2,7 +2,7 @@
 try:
     import MeCab
     import ipadic
-except ModuleNotFoundError:
+except ImportError:
     # Don't fail until the tokenizer is actually used
     MeCab = None
 


ozancaglayan:
Thanks for catching this. Do we consider this urgent enough to do another release @mjpost

mjpost:
Sure, that's fine. But I suggest adding a Python 3.5 profile to the regression testing first. Once we have that and it's failing, then merge this in, and we can re-release.

(I'd actually love to get rid of Python 3.5 support, so we can have f-strings, but I think we'll wait for that).

ozancaglayan:
Do you mean travis? If so, there is already a 3.5 target there but this did not get caught because mecab should be missing for the exception to be triggered (i guess?)

mjpost:
Is there any easy way to test that scenario?

ozancaglayan:
Humm, no idea. Why do you want this particular condition to be caught by Travis before merging this?

thomasZen:
> Yes, I meant travis. Is there any easy way to test that scenario?

@mjpost @ozancaglayan I didn't check the CI workflow of this repository but it's easily possible to add that to a static type checker (which is useful to have for a repository that contains type annotations).

Here's an example file `foo.py`:
```
try:
    import fooBar  # type: ignore
except ModuleNotFoundError:
    pass
```

For example if you use [mypy](http://mypy-lang.org/) and run `mypy --python-version 3.6 foo.py` it will pass. With`mypy --python-version 3.5 foo.py ` you'll get the following error:
`foo.py:3: error: Name 'ModuleNotFoundError' is not defined`

A other option is to just run sacrebleu with python 3.5 without installing the japenese tokenizer. This will immediately fail once the file that uses `ModuleNotFoundError` is imported.

mjpost:
Hi @ozancaglayan, the idea is just that it's good practice to test for things that have caused active bugs.

This bug isn't likely to happen again, it was just a suggestion.

ozancaglayan:
I think this requires integrating `mypy` as suggested by @thomasZen . I would go with merging this and later we can add this kind of testing.

mjpost:
Done (though you should feel free to do this yourself, too).

PR #113, opened 2020-08-11
Added char-based tokenization

mjpost:


mjpost:
This looks complicated but is just merging in #93.

Issue #121, opened 2020-10-22
Sentence CHRF silently accepts single reference as string


bricksdont:
Hi Matt, all

While trying to debug an issue related to sentence-level CHRF, I noticed that the compat function `sentence_chrf` expects a list of references as input, but happily accepts a single string (a list of characters). But then only uses the first character as the reference:

```
>>> from sacrebleu import sentence_chrf
>>> sentence_chrf("hi there", "hi there")
chrF2 = 0.455
>>> sentence_chrf("hi there", "h")
chrF2 = 0.455
```

While the `sentence_score` function for BLEU intercepts strings instead of lists of strings:

https://github.com/mjpost/sacrebleu/blob/65747b6ef6846d49548fd5139d2a6d118c7a384e/sacrebleu/metrics/bleu.py#L227

the one for CHRF does not:

https://github.com/mjpost/sacrebleu/blob/65747b6ef6846d49548fd5139d2a6d118c7a384e/sacrebleu/metrics/chrf.py#L125

Could this perhaps be changed / standardized for all metrics?

ozancaglayan:
Thanks for reporting this. The PR synchronizes the behaviors across metrics.

bricksdont:
A similar issue, I think, is that corpus CHRF expects `List[Iterable[str]]` as input, and silently accepts `List[str]` which will fail in `zip_longest`, but the error message will be confusing.

https://github.com/mjpost/sacrebleu/blob/65747b6ef6846d49548fd5139d2a6d118c7a384e/sacrebleu/metrics/chrf.py#L136

ozancaglayan:
I did not understand the latter issue. The signature of `corpus_*` is the same both in BLEU and chrF. I accept that having a compat API and also object-oriented API at the same time is kind of confusing, and the latter should be supported with more examples/docs and maybe unit tests to discover potential flaws & argument inconsistencies.

bricksdont:
In this function, there is already code to "add some robustness to the input arguments":

https://github.com/mjpost/sacrebleu/blob/65747b6ef6846d49548fd5139d2a6d118c7a384e/sacrebleu/metrics/chrf.py#L146

The code intercepts strings as inputs to this function and makes them into lists. Could it also intercept / give an error message if the references argument is a simple list `List[str]` instead of the expected `List[List[str]]`?

[...]

PR #122, opened 2020-10-22
chrf: assert if refs are not in a list (#121)

ozancaglayan:
Bail out if references are not given in a list, for `chrf.sentence_score()` method.

### DIFF
--- a/sacrebleu/metrics/bleu.py
+++ b/sacrebleu/metrics/bleu.py
@@ -224,7 +224,8 @@ class BLEU:
         :param use_effective_order: Account for references that are shorter than the largest n-gram.
         :return: a `BLEUScore` object containing everything you'd want
         """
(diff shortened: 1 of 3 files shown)


martinpopel (approved):
Thanks.

bricksdont:
Bumping this issue; the build for Python 3.5 fails, I think because mecab-python3 is not compatible with Python 3.5

Other than that, I think that this PR is straightforward and could be merged - thanks!

ozancaglayan:
yeah I noticed that one as well.

ozancaglayan:
@polm can we pin to mecab 1.0.2 here, so that we're still supporting python 3.5 (until we decide whether we drop support for 3.5 for sacreBLEU as well) ?

ozancaglayan:
Thanks @polm , pinning to 1.0.3 seemed to fix it

Issue #39, opened 2019-07-19
The Choice of Detokenization Methods


jcyk:
Hi, 
Just a couple of questions since we need to detokenize data before evaluation.
Will the choice of detokenization method affect the BLEU score?
Which detokenization method is the best choice?

mjpost:
Yes, the detokenization and normalization will have some effect. The general answer is that you should detokenize so as to be closest to the form the reference takes. One issue that doesn’t come out in papers often is that of normalization. For example, if the references uses smart quotes, but your output uses straight quotes, it can have a large effect on the BLEU score, if your corpus has a lot of quotations.

martinpopel:
A related question for @mjpost: Would you welcome a PR with `--no-punct` option, which deletes all punctuation (from both the reference and translation)? As for the signature, it could either add "no-punct" to the tokenization method, or introduce "+punct.no" (but I don't want to make the default signature any longer). This option could be helpful when detecting e.g the effect of smart/straight quotes. It is similar to lowercasing in its nature.

mjpost:
Yes, this sounds great. There’s a great universal implementation from @arya.mccarthy we could borrow from here:

    https://github.com/duolingo/duolingo-sharedtask-2020/blob/a000131a8714a6832069cfc0b281a187e8ed3eb6/utils.py#L51

(agree about keeping default sig shorter. Wish I’d put more thought into it (e.g., invertible Geek codes)

matt (from my phone)

PR #123, opened 2020-10-30
adds tokenizer option to remove punctuation (except apostrophes)

esalesky:
Adds a tokenizer to remove punctuation (except apostrophes). This is to enable sacrebleu to be used for ASR corpora which have been translated (specifically Fisher Spanish--English) where the convention is to compute BLEU on lowercased, no-punct-except-apostrophes text (for better or for worse)

### DIFF
--- a/sacrebleu/dataset.py
+++ b/sacrebleu/dataset.py
@@ -22,6 +22,42 @@
 # Many of these are *.sgm files, which are processed to produced plain text that can be used by this script.
 # The canonical location of unpacked, processed data is $SACREBLEU_DIR/$TEST/$SOURCE-$TARGET.{$SOURCE,$TARGET}
 DATASETS = {
+    "mtedx/valid": {
+        "data": ['https://raw.githubusercontent.com/esalesky/mtedx-eval/main/valid.tar.gz'],
+        "description": 'mTEDx evaluation data, valid',
+        "md5": ['866e44c55657c6c485ef96a244c0fe07'],
+        "el-en": ['valid/mtedx-valid-elen.el','valid/mtedx-valid-elen.en'],
+        "es-en": ['valid/mtedx-valid-esen.es','valid/mtedx-valid-esen.en'],
+        "es-fr": ['valid/mtedx-valid-esfr.es','valid/mtedx-valid-esfr.fr'],
+        "es-it": ['valid/mtedx-valid-esit.es','valid/mtedx-valid-esit.it'],
+        "es-pt": ['valid/mtedx-valid-espt.es','valid/mtedx-valid-espt.pt'],
+        "fr-en": ['valid/mtedx-valid-fren.fr','valid/mtedx-valid-fren.en'],
+        "fr-es": ['valid/mtedx-valid-fres.fr','valid/mtedx-valid-fres.es'],
+        "fr-pt": ['valid/mtedx-valid-frpt.fr','valid/mtedx-valid-frpt.pt'],
+        "it-en": ['valid/mtedx-valid-iten.it','valid/mtedx-valid-iten.en'],
+        "it-es": ['valid/mtedx-valid-ites.it','valid/mtedx-valid-ites.es'],
+        "pt-en": ['valid/mtedx-valid-pten.pt','valid/mtedx-valid-pten.en'],
+        "pt-es": ['valid/mtedx-valid-ptes.pt','valid/mtedx-valid-ptes.es'],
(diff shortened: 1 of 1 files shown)


martinpopel:
I had a similar idea for a long time and even a work-in-progress commit: https://github.com/martinpopel/sacreBLEU/commit/cb0227934bb256d3b6e62e698b37fcb19c0d8023
So thanks for reminding me about this.

While stripping punctuation seems simple enough, there subtleties, which should be well designed before everyone starts using this option and it will be difficult to change the semantics.

* You should add a test (e.g. adding one line to `test.sh` as I did).
* When using the third-party `regex` module you should add it as a dependency to `setup.py` (I guess Travis CI should report this). An alternative is to use ony Python-core `unicodedata` and `@functools.lru_cache(maxsize=1)` (again, see my solution).
* The `nopea` name does not sound very intuitive to me. Even if we agree that apostrophes should be kept (for compatibility with legacy ASR conventions), I think the name should be no-punct.
* While the preferred Unicode code point for apostrophe is U+2019 (’), the good-old U+0027 is being used as well. Both symbols are being used for single quotation (the former as the RIGHT SINGLE QUOTATION MARK). An unfortunate implication of your decision to keep U+0027, but delete U+2019 (’) is that the tokenizer cannot be used for [ignoring the difference between smart quotes and straight quotes](https://github.com/mjpost/sacrebleu/issues/39#issuecomment-617226768). Of course, we can have a separate option/tokenizer for that purpose, but too many options with similar purpose may re-introduce the problems sacrebleu is trying to solve.
* It is a question if `no-punct` should be implemented as a tokenizer or a pre-processing (or post-processing) option similar to `-lc` for lowercasing. The latter can be combined with other tokenizers and I was not able to decide if it is an advantage or a disadvantage.
* There is a question if we want to delete decimal-point markers (and thousand separators), i.e. commas and dots withing numbers. Note that these punctuation marks are not tokenized by the default `13a` tokenizer (as well as the `intl`).

I like that `-lc` gives always a higher-or-equal BLEU score relative to the default case-sensitive version. This way I can quickly asses how many errors could be fixed if the hypothesis gets all casing correctly. We can interpret such case-insensitive BLEU as a kind of oracle score. I would love to have a similar mechanism for punctuation. However, if you delete all punctuation it affects the tokenization, so for example "a.m." becomes a single token "am" instead of four tokens. A better alternative seems to be to substitute punctuation with space (and merge multiple spaces into one, as you do, although it is [not needed](https://github.com/mjpost/sacrebleu/blob/65747b6ef6846d49548fd5139d2a6d118c7a384e/sacrebleu/metrics/bleu.py#L100)). Unfortunately, this still does not achieve my goal: punctuation is usually very easy to translate, so by deleting all punctuation tokens, we usually decrease the final score. My next attempt was to convert all punctuation symbols to a single symbol, e.g. a dot. This way the number of tokens (and thus also the brevity penalty) would remain the same. Similarly to `-lc`, it could not make BLEU lower. It would serve as a crude punctuation normalization, but it would not work for your purpose of ASR output without punctuation and references with punctuation. It would be also misleading to call this `no-punct` or `ignore-punct`. This was the reason why I did not finish my PR. My current suggestion is to implement both `no-punct` (either as a tokenizer or a pre-processing option) and `--normalize-punct` (either as [in Moses](https://github.com/moses-smt/mosesdecoder/blob/master/scripts/tokenizer/normalize-punctuation.perl) or by dully mapping all punctuation to dots). These should be two separate PRs, of course.

esalesky:
Thank you for the very good points!! 

I agree with your suggestion of having both `--no-punct` and `--normalize-punct`. Your motivation of normalizing punctuation and having a check for the resulting difference if punctuation is correctly treated in text-based MT is a good one, but `normalize-punct` in that way wouldn't enable use for the Fisher Spanish--English corpus convention. Having both would I think hit both goals without needing more options. 

Renaming `nopea` to either `no-punct` or `speech` seems good to me, and I will add a test to test.sh
I implemented it as a tokenizer to make it easy to combine (or not) with `--lc`, and because for my motivation of translated speech corpora, it wouldn't be combined with another tokenizer. For `normalize-punct` a pre or post-processing option seems right to me. 

A question: I can add regex to setup.py or switch to unicodedata; re is currently imported in setup.py, but its behavior with escaping chars changed in python 3.6 and 3.7 (hence my use of import regex as re). Would there be a reason not to switch to import regex from re in setup.py? I don't believe it should require changes elsewhere.

esalesky:
[closing to submit another request, but I will come back to this!]

Issue #130, opened 2021-01-08
BLEU with a variable number of references


tuetschek:
The current implementation of BLEU in SacreBLEU doesn't allow a variable number of references (different segments having a different number of references). Some NLG datasets (e.g. [E2E](https://github.com/tuetschek/e2e-dataset), [WebNLG](https://gitlab.com/shimorina/webnlg-dataset)) do have a variable number of references and it would be great if SacreBLEU worked on them.

Is that a design decision, or would you be open to a PR that would enable this?

martinpopel:
It seems that this feature is often needed (e.g. #69) and the current workaround (duplicate another reference for a given segment) is not user friendly (and it is not efficient), so I would welcome such a PR.

### File format
Currently, sacreBLEU supports both a single file with tab-separated references and multiple files (each reference in a single file).

The former format currently needs `--num-refs=Integer`. So do we want to allow also a range, e.g. `--num-refs=1-5`, so there is still same sanity check of the input data?

Do we want to allow variable number of references also in the latter format? That would mean e.g. re-defining an empty string as a missing reference (which should not affect the brevity penalty). How would then someone represent a real empty segment as a reference? (Suppose there is a language where the best translation of "Um, heh." is to omit it.) We could disambiguate the two use cases by special options such as `--variable-number-of-references` or `--treat-empty-as-empty`, but that seems too complicated (even with better names).

### The Python API
Currently, we have `corpus_bleu(sys, [ref1, ref2,...refN])`, where each of the N references is a list of all segments. I am not sure if we want to change it (cf. #125).
We could represent missing references (for a given segment) with `None` here, preventing thus the ambiguity of empty strings.

tuetschek:
Thanks for such a quick and detailed answer! I'll try to implement it then 🙂! 

I was actually only considering the Python API so far (for the upcoming [GEM NLG shared task/benchmark](https://gem-benchmark.github.io/)), so having the possibility just via the API and not exposed on the command line would be fine for my use case. I agree that representing missing references with `None` is the easiest way to go.

Re. Command line/file format: I think that allowing variable numbers of references just in the tab-separated format is fine. Allowing a range for `--num-refs` makes sense.

Regardless of whether we want to support the multi-file format or not, would it make sense to represent real empty references with a space character instead of a truly empty line/tab column, or is that too obscure?

ozancaglayan:
Hello

Thanks for the issue and the PR!

[...]

PR #132, opened 2021-01-12
Allow variable number of references for BLEU via API

tuetschek:
This allows BLEU to use a variable number of references (different number of references for each sentence) if called via the API (see #130). If a sentence has fewer than the maximum number of references, `None` should used to fill remaining reference streams.

### DIFF
--- a/sacrebleu/metrics/bleu.py
+++ b/sacrebleu/metrics/bleu.py
@@ -281,7 +281,7 @@ class BLEU:
         fhs = [sys_stream] + ref_streams
         for lines in zip(*fhs):
             # remove undefined references (i.e. we have fewer references for this particular sentence)
-            lines = [x for x in lines if x is not None]
+            lines = [x for x in lines if x is not None and x != ""]
             if len(lines) < 2:  # we need at least system + 1 defined reference
                 raise EOFError("No valid references for a sentence!")
 
--- a/sacrebleu/metrics/bleu.py
+++ b/sacrebleu/metrics/bleu.py
@@ -3,7 +3,6 @@
 import math
 import logging
 from collections import Counter
-from itertools import zip_longest
 from typing import List, Iterable, Union
 
 from ..tokenizers import TOKENIZERS
@@ -256,10 +255,18 @@ class BLEU:
         # look for already-tokenized sentences
         tokenized_count = 0
 
+        # sanity checks
+        if any(len(ref_stream) != len(sys_stream) for ref_stream in ref_streams):
+            raise EOFError("System and reference streams have different lengths!")
+        if any(line is None for line in sys_stream):
+            raise EOFError("Undefined line in system stream!")
+
         fhs = [sys_stream] + ref_streams
-        for lines in zip_longest(*fhs):
-            if None in lines:
-                raise EOFError("Source and reference streams have different lengths!")
+        for lines in zip(*fhs):
+            # remove undefined references (i.e. we have fewer references for this particular sentence)
+            lines = [x for x in lines if x is not None]
+            if len(lines) < 2:  # we need at least system + 1 defined reference
+                raise EOFError("No valid references for a sentence!")
 
             if self.lc:
                 lines = [x.lower() for x in lines]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
     to make two methods behave the same.
   - Add smoothing value to BLEU signature (#98)
   - dataset: Fix IWSLT links (#128)
+  - Allow variable number of references for BLEU (only via API) (#130)
 
 - 1.4.14 (2020-09-13)
   - Added character-based tokenization (`-tok char`).


martinpopel (approved):
Nice, thanks.
Let's merge this and the command-line support for variable number of refs can be solved in another PR (if someone needs it).

sacrebleu/metrics/bleu.py
```diff
@@ -256,10 +255,18 @@ def corpus_score(self, sys_stream: Union[str, Iterable[str]],
...
+            raise EOFError("Undefined line in system stream!")
+
         fhs = [sys_stream] + ref_streams
-        for lines in zip_longest(*fhs):
-            if None in lines:
-                raise EOFError("Source and reference streams have different lengths!")
+        for lines in zip(*fhs):
+            # remove undefined references (i.e. we have fewer references for this particular sentence)
+            lines = [x for x in lines if x is not None]
+            if len(lines) < 2:  # we need at least system + 1 defined reference
```
mjpost, line 268: What if we change this to

```python
lines = [x for x in lines if x is not None and x != ""]
```

Then the command line would work fine, unless I am missing something. You could, for example, provide 100 reference files, and if there were not 100 references for sentence 10, you could just have blank lines, which would then be skipped.

mjpost:
Thanks for this! I suggest we merge into #133 and including it with the changes to version 1.5.

sacrebleu/metrics/bleu.py
martinpopel: This PR solves just the API, not the command-line interface and file-format.
My idea was (see https://github.com/mjpost/sacrebleu/issues/130#issuecomment-757051390) that
* In the case of tab-separated refs in a single file, we would just specify a range of the number of references (`--num-refs=1-5`), but no empty strings would be needed. Missing columns would be substituted with `None` automatically before calling the API.
* In the case of multiple references in multiple files, we would need empty lines for missing references, but these would be again converted to `None` before calling the API.

The advantage is that the API could be still used by someone who needs empty references for some segments. This is unusual in MT, but BLEU (and `sacrebleu`) may be used also in other tasks.

That said, I don't have strong opinion on this. If we add `and x != ""` now, the feature will be immediately available also from the command line and no further PRs will be needed (just a documentation).

ozancaglayan: (FYI: `lines = [x for x in lines if x]` should be enough to test for both emptyness and None)

ozancaglayan: Should we merge this as it is then? What's the final decision?

martinpopel: Both merging as is (handling `None` only) and @mjpost's suggesting for handling also empty strings are OK for me.
I would slightly prefer the former as it allows empty references for some tasks (but I am not aware of any such task).

> `lines = [x for x in lines if x]` should be enough to test for both emptyness and None

Yes, but I find the explicit `if x is not None and x != ""` better because it is self-documenting that these are the two expected use cases (and `False`, empty containers and numerical zero are not expected).

mjpost: As @martinpopel suggests, I like the more explicit version.

mjpost:
I'm happy to have this merged, but it needs a note added to the CHANGELOG.

sacrebleu/metrics/bleu.py
ozancaglayan: Just after merging this, I figured out that this also removes empty hypotheses, which could have been produced by a decoder that only generated <EOS> token in the early training phase. In the case of multi-references, first ref would then be taken as the `hyp`, leading to erroneous computation:

```python
In [22]: sys
Out[22]: 
['The dog bit the man.',
 "It wasn't surprising.",
 'The man had just bitten him.']

In [23]: refs
Out[23]: 
[['The dog bit the man.', 'It was not unexpected.', 'The man bit him first.'],
 ['The dog had bit the man.',
  'No one was surprised.',
  'The man had bitten the dog.']]

In [24]: empty_sys
Out[24]: ['', '', '']

In [25]: sacrebleu.corpus_bleu(sys, refs)
Out[25]: BLEU = 48.53 82.4/50.0/45.5/37.5 (BP = 0.943 ratio = 0.944 hyp_len = 17 ref_len = 18)

In [26]: sacrebleu.corpus_bleu(empty_sys, refs)
Out[26]: BLEU = 23.93 64.7/35.7/18.2/12.5 (BP = 0.889 ratio = 0.895 hyp_len = 17 ref_len = 19)
```

ozancaglayan:
Just after merging this, I figured out that this also removes empty hypotheses, which could have been produced by a decoder that only generated "EOS" token in the early training phase. In the case of multi-references, first ref would then be taken as the `hyp`, leading to erroneous computation:

```python
In [22]: sys
Out[22]: 
['The dog bit the man.',
 "It wasn't surprising.",
 'The man had just bitten him.']

In [23]: refs
Out[23]: 
[['The dog bit the man.', 'It was not unexpected.', 'The man bit him first.'],
 ['The dog had bit the man.',
  'No one was surprised.',
  'The man had bitten the dog.']]

In [24]: empty_sys
Out[24]: ['', '', '']

In [25]: sacrebleu.corpus_bleu(sys, refs)
Out[25]: BLEU = 48.53 82.4/50.0/45.5/37.5 (BP = 0.943 ratio = 0.944 hyp_len = 17 ref_len = 18)

In [26]: sacrebleu.corpus_bleu(empty_sys, refs)
Out[26]: BLEU = 23.93 64.7/35.7/18.2/12.5 (BP = 0.889 ratio = 0.895 hyp_len = 17 ref_len = 19)
```

sacrebleu/metrics/bleu.py
tuetschek: @ozancaglayan : Good point. There's a sanity check for non-null hypotheses on line 261, but I guess that we need to allow empty hypotheses, so you can't just add `!= ''` there.

This should work:
```
lines = [lines[0]] + [x for x in lines[1:] if x is not None and x != ""]
```
But it's not pretty. Maybe it would be easier to treat the hypothesis separately from the references? I guess they're only joined together for the lowercasing and the tokenizer? 

Let me know, I can push a fix.

ozancaglayan: Iterating over them separately would be more cleaner but probably it's done this way for tokenization and lowercasing, as they are splitted back into different iterators in line 299.

Your fix seems legitimate to me, care to add 1-2 comment lines there to explain it as well?

I hope one day, we'll refactor these I/O pre-processing stuff out of the core computation methods, which would make things much clean.

mjpost: Very good catch! Did we have a test case that failed that triggered this? If not, we really should.

Can we do something like

```python
for sys, *refs in zip(*fhs):
   lines = [sys] + [ref for ref in refs if ref is not None and ref != ""]
```

That is also a bit ugly but at least the meaning is explicit.

tuetschek: OK, I've created #134 with the fix – I hope it's OK, I'm not sure how to add it here after this has been merged.

PR #133, opened 2021-01-12
Some fixes staged for 1.5

ozancaglayan:
Hello,

I always end up doing this in a sub-optimal way. This one includes stuff that should normally be created as different PRs probably. Maybe we can merge without squashing this time?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,14 @@
 # VERSION HISTORY
 
-- 1.5 (in progress)
+- 1.5.0 (2021-01-15)
   - Fix an assertion error in chrF (#121)
   - Add missing `__repr__()` methods for BLEU and TER
   - TER: Fix exception when `--short` is used (#131)
   - Pin Mecab version to 1.0.3 for Python 3.5 support
   - [API Change]: Default value for `floor` smoothing is now 0.1 instead of 0.
-  - [API Change]: `sacrebleu.sentence_bleu()` now uses the `exp` smoothing method
-    similar to the CLI's --sentence-level behavior.
+  - [API Change]: `sacrebleu.sentence_bleu()` now uses the `exp` smoothing method,
+    exactly the same as the CLI's --sentence-level behavior. This was mainly done
+    to make two methods behave the same.
   - Add smoothing value to BLEU signature (#98)
   - dataset: Fix IWSLT links (#128)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # VERSION HISTORY
 
 - 1.4.15 (in progress)
+  - Fix an assertion error in chrF (#121)
+  - Add missing `__repr__()` methods for BLEU and TER
+  - TER: Fix exception when `--short` is used (#131)
+  - Pin Mecab version to 1.0.3 for Python 3.5 support
 
 - 1.4.14 (2020-09-13)
   - Added character-based tokenization (`-tok char`).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
   - [API Change]: `sacrebleu.sentence_bleu()` now uses the `exp` smoothing method
     similar to the CLI's --sentence-level behavior.
   - Add smoothing value to BLEU signature (#98)
+  - dataset: Fix IWSLT links (#128)
 
 - 1.4.14 (2020-09-13)
   - Added character-based tokenization (`-tok char`).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,14 @@
 # VERSION HISTORY
 
-- 1.4.15 (in progress)
+- 1.5 (in progress)
   - Fix an assertion error in chrF (#121)
   - Add missing `__repr__()` methods for BLEU and TER
   - TER: Fix exception when `--short` is used (#131)
   - Pin Mecab version to 1.0.3 for Python 3.5 support
+  - [API Change]: Default value for `floor` smoothing is now 0.1 instead of 0.
+  - [API Change]: `sacrebleu.sentence_bleu()` now uses the `exp` smoothing method
+    similar to the CLI's --sentence-level behavior.
+  - Add smoothing value to BLEU signature (#98)
 
 - 1.4.14 (2020-09-13)
   - Added character-based tokenization (`-tok char`).
--- a/sacrebleu/__init__.py
+++ b/sacrebleu/__init__.py
@@ -14,7 +14,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '1.4.14'
+__version__ = '1.5'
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 # VERSION HISTORY
 
-- 1.4.15 (in progress)
+- 1.5 (in progress)
+  - Fix an assertion error in chrF (#121)
```
mjpost, line 4: Make sure to change this to the data (2021-01-12 is fine even if we don't end up approving today)

CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
...
+  - Add missing `__repr__()` methods for BLEU and TER
+  - TER: Fix exception when `--short` is used (#131)
+  - Pin Mecab version to 1.0.3 for Python 3.5 support
+  - [API Change]: Default value for `floor` smoothing is now 0.1 instead of 0.
+  - [API Change]: `sacrebleu.sentence_bleu()` now uses the `exp` smoothing method
+    similar to the CLI's --sentence-level behavior.
+  - Add smoothing value to BLEU signature (#98)
```
mjpost, line 11: Not just "similar", but the same! Maybe worth noting this was done to bring make their behavior the same

sacrebleu/__init__.py
```diff
@@ -14,7 +14,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-__version__ = '1.4.14'
+__version__ = '1.5'
 __description__ = 'Hassle-free computation of shareable, comparable, and reproducible BLEU, chrF, and TER scores'
```
mjpost, line 18: Should this be 1.5.0? Or maybe trailing zeros are left off, not sure.

martinpopel (approved):
Great. Thanks. Just a minor question/suggestion on moving `__repr__` to `BaseScorer`.

sacrebleu/compat.py
```diff
@@ -40,13 +38,14 @@ def corpus_bleu(sys_stream: Union[str, Iterable[str]],
 
 def raw_corpus_bleu(sys_stream,
                     ref_streams,
-                    smooth_value=None) -> BLEUScore:
+                    smooth_value=0) -> BLEUScore:
     """Convenience function that wraps corpus_bleu().
```
mjpost, line 42: I think we should factor out the 0 here into a global constant, e.g., `SMOOTH_VALUE_DEFAULT`.

sacrebleu/metrics/bleu.py
```diff
@@ -62,15 +75,22 @@ def format(self, width=2, score_only=False, signature=''):
...
     SMOOTH_DEFAULTS = {
-        'floor': 0.0,
+        # The defaults for `floor` and `add-k` are obtained from the following paper
+        # A Systematic Comparison of Smoothing Techniques for Sentence-Level BLEU
+        # Boxing Chen and Colin Cherry
+        # http://aclweb.org/anthology/W14-3346
+        'none': None,   # No value is required
+        'floor': 0.1,
         'add-k': 1,
         'exp': None,    # No value is required
```
mjpost, line 93: These, too, should maybe have global constants, so we don't inadvertently use one in only one place?

mjpost (commented):
LGTM, pending the small comments.

sacrebleu/compat.py
ozancaglayan: The defaults for BLEU are already defined in `BLEU.SMOOTH_DEFAULTS`. The reason it is hard-coded to be 0 here was to preserve backward compatibility with the `raw_corpus_bleu` of older sacreBLEU versions. But since we're breaking API here, I can revert this to be `None` so that it defaults to the new internal default of 0.1 ?

sacrebleu/metrics/bleu.py
ozancaglayan: Can't we consider these as global constants since this is a class variable dictionary named `BLEU.SMOOTH_DEFAULTS` ?

sacrebleu/metrics/bleu.py
```diff
@@ -62,6 +62,9 @@ def format(self, width=2, score_only=False, signature=''):
             rl=self.ref_len)
         return s
 
+    def __repr__(self):
+        return self.format()
+
```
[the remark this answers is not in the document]
ozancaglayan: Done.

CHANGELOG.md
ozancaglayan: done

CHANGELOG.md
ozancaglayan: Done, changed to 15 January.

sacrebleu/__init__.py
ozancaglayan: done

sacrebleu/compat.py
ozancaglayan: The alternative is to remove this "convenience" function completely and expect people to use corpus_bleu with arguments that they'd like to use.

mjpost: So `raw_corpus_bleu` hard-codes a `smooth_value` of `floor`. Since the API allows the value to be changed, I would suggest that instead of `smooth_value=0`, we use `smooth_value=BLEU.SMOOTH_DEFAULTS["floor"]`. That way the value is explicit, instead of potentially getting lost here. What do you think of this?

sacrebleu/metrics/bleu.py
mjpost: Oh, sorry, I missed what file this was in (Github cut off the top line showing that this is the actual definition of `SMOOTH_DEFAULTS`)! So ignore my comment here.

sacrebleu/compat.py
mjpost: I'd like to keep the convenience function. I think Sockeye might be using it.

mjpost:
I think I'm happy with merging this, subject to the few stray comments. Does one of you want to handle this, or do you prefer I do it? I think we should do a squash merge per our earlier conversation, but take care to create a tidy commit message.

sacrebleu/compat.py
ozancaglayan: `smooth_value=BLEU.SMOOTH_DEFAULTS["floor"]` is equivalent (although more explicit) to the unmodified version of the signature which was `smooth_value=None` (a `None` fetches the default from that very same dictionary internally in `Bleu` class)

The question here is whether we want sacreBLEU >= 1.5 to give the same sentence bleu scores as sacreBLEU < 1.5 when using this function, or not. I think we don't care as of now since we already introduced other behavior differences as well.

So I'll make this `BLEU.SMOOTH_DEFAULTS["floor"]` to be more readable.

The reason why I suggested this function is that, it again falls back to `floor` although all other methods for sentence-bleu are now synchronized to use `exp` smoothing. It's kind of weird the choices here =)

ozancaglayan:
I don't know how to merge the other variable-bleu PR into this. If you want, we can merge this one first, and then rebase the other PR, add the Changelog entry and merge afterwards?

mjpost:
Sure, that's fine. It doesn't particularly matter to me.

PR #134, opened 2021-01-18
BLEU variable references: fix for empty hypotheses

tuetschek:
This is a fix for a problem @ozancaglayan mentioned in #132 -- if the hypothesis is empty, we don't want it to be removed from `lines` and compare just references among themselves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@
     to make two methods behave the same.
   - Add smoothing value to BLEU signature (#98)
   - dataset: Fix IWSLT links (#128)
-  - Allow variable number of references for BLEU (only via API) (#130)
+  - Allow variable number of references for BLEU (only via API) (#130).
+    Thanks to Ondrej Dusek (@tuetschek)
 
 - 1.4.14 (2020-09-13)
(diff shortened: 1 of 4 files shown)


ozancaglayan:
@mjpost I've added some new tests here, including the one for this specific case.

ozancaglayan (approved):

mjpost:
Thanks, all! I just released [sacrebleu 1.5.0](https://pypi.org/project/sacrebleu/).

ozancaglayan:
thanks, sorry didn't have time for the final thing. Are you planning to do a release on github as well? Because it's still showing 1.4.13 as the latest release there.

mjpost:
No problem. It's in the Makefile, though: `make publish`.

I have never done a release manually. I thought Github did it automatically, but I see you did the last one yourself. I just went through the process manually, though. I wonder if, when I created the tag, I should have used `git tag -a` and then added the bit from the Changelog. I suspect that would have auto-created a release.

ozancaglayan:
Thinking back on this, although this is only available through the API, the `num_refs` in the signature became quite ambiguous, no?

PR #151, opened 2021-03-26
link to Papineni et al. (2002)

martinpopel:
Sacrebleu is being used by users not familiar with the definition of BLEU.
We should at least give a link (and credit) to the original BLEU paper.


ozancaglayan:
Hi,

This is already done in my 2.0.0 branch for which I'll send a PR soon.

martinpopel:
OK.

Issue #144, opened 2021-03-03
chrF not compatible with chrF++, Moses and NLTK for sentence-level smoothing


ozancaglayan:
Created a new issue for this problem, to detach it from #121
See the comment: https://github.com/mjpost/sacrebleu/issues/121#issuecomment-786608571

To summarize, sacreBLEU applies effective order smoothing whereas others smooth with small value `eps`. The fix is already in my branch for sacreBLEU 2.0.0. The problem is to agree on whether we accept changing behavior for sentence-level chrF for >=2.0.0.

martinpopel:
Yes, I vote for being compatible with the chrF++ implementation of chrF by default. We can keep the effective-order smoothing as a non-default option, or drop it (for chrF).

Issue #141, opened 2021-02-25
Sentence BLEU yields non-intuitive scores


ozancaglayan:
I noticed this when I intended to replace `nltk`'s sentence bleu with sacreBLEU in some of my code. We should probably just return 0 if there are no matches at all.

```

In [55]: sacrebleu.sentence_bleu('yes', ['no'], smooth_method='exp')
Out[55]: BLEU = 50.00 50.0/0.0/0.0/0.0 (BP = 1.000 ratio = 1.000 hyp_len = 1 ref_len = 1)

In [56]: sacrebleu.sentence_bleu('yes', ['no'], smooth_method='floor')
Out[56]: BLEU = 10.00 10.0/0.0/0.0/0.0 (BP = 1.000 ratio = 1.000 hyp_len = 1 ref_len = 1)

In [46]: hyp
Out[46]: 'this is a cat'

In [47]: ref
Out[47]: 'okay thanks'

In [48]: sacrebleu.sentence_bleu(hyp, [ref], smooth_method='floor')
Out[48]: BLEU = 4.52 2.5/3.3/5.0/10.0 (BP = 1.000 ratio = 2.000 hyp_len = 4 ref_len = 2)

In [49]: sacrebleu.sentence_bleu(hyp, [ref], smooth_method='exp')
Out[49]: BLEU = 7.99 12.5/8.3/6.2/6.2 (BP = 1.000 ratio = 2.000 hyp_len = 4 ref_len = 2)

% nltk
In [50]: sentence_bleu([ref.split()], hyp.split(), smoothing_function=SmoothingFunction().method3)
Out[50]: 0

In [51]: sentence_bleu([ref.split()], hyp.split(), smoothing_function=SmoothingFunction().method2)
Out[51]: 0
```

ozancaglayan:
There's actually a test that assumes non-zero scores will be produced although there are no matches between hyps & refs:

```
test_raw_bleu_cases = [                                                            
    (["this is a test", "another test"], [["ref1", "ref2"]], 0.003799178428257963),
   ...       
```

Do we all agree that this is a bug that should be fixed @mjpost @martinpopel ?

martinpopel:
I agree that if there is no match even within unigrams, BLEU should be zero, otherwise it is not intuitive.

Issue #140, opened 2021-02-24
Scaling chrF and TER to 0-100


ozancaglayan:
At least for chrF, the original implementation reports the score that way. Do you think it is reasonable to standardize these across the three metrics? It'll also make reportin a bit more consistent w.r.t to the `--width` parameter.

martinpopel:
Yes, I would love that.
I think it is OK for everyone, especially if it will be in sacreBLEU 2.0, where such changes are expected.

cfedermann:
+1

Issue #124, opened 2020-11-24
Having better defaults for ChrF


alvations:
The chrf++ is a better chrf than the default chrf6 used in sacrebleu, could we change the default to chrf++?

From the original ChrF creator: https://twitter.com/amelija16mp/status/1331288013880614913

ozancaglayan:
This would require changing `chrf_order` param to `chrf_char_order`, and adding a new `chrf_word_order` argument. For Chrf++, I think the defaults are 6 and 2, respectively. The current chrF in sacreBLEU probably amounts to these params being 6 and 0. Can someone confirm?

alvations:
@m-popovic should be the "authoritative" person to confirm that =)

ozancaglayan:
I added this into my `refactor2021` branch. I should note that chrF++.py does a primitive tokenization for word-level n-gram matching e.g. it only separates out punctuations from beginning and end of words. Can someone come up with a regexp to replace the below implementation that I've got from chrF++?

```
def separate_punctuation(line):
    words = line.strip().split()
    tokenized = []
    for w in words:
        if len(w) == 1:
            tokenized.append(w)
        else:
            lastChar = w[-1] 
            firstChar = w[0]
            if lastChar in string.punctuation:
                tokenized += [w[:-1], lastChar]
            elif firstChar in string.punctuation:
                tokenized += [firstChar, w[1:]]
            else:
                tokenized.append(w)
    
    return tokenized
```

Also, because of the way the sentences are tokenized as above, languages like zh and ja would probably not benefit from having word-level ngrams, right?

martinpopel:
> it only separates out punctuations from beginning and end of words

Yes, only a single punctuation character, only at beginning and end of words, but if there is a punctuation at the end, no punctuation from the beginning is separated:
```python
separate_punctuation('(hi)') == ['(hi', ')']
```
This seems very wrong, but if we need to replicate chrF++, we need to accept that.

> I am curious about the regexp as well

The following code is so obfuscated and also about 3.3 times slower than the original, so I mention it just as a joke:
```python
import string
punct = string.punctuation
ugly_re = f'\s+|((?<=\S)[{punct}](?=\s|$)|(?<=\s)[{punct}](?=\S*[^{punct}](?:\s|$)))'

def separate_punctuation_re(line):
    return list(filter(None, re.split(ugly_re, line)))
```

The following code seems to be 23% faster than the original (on a randomly chosen file):
```python
def separate_punctuation4(line):
[...]

PR #152, opened 2021-03-26
Changes for 2.0.0

ozancaglayan:
Hello,

- Here's a detailed summary of this PR. It'll probably be quite hard to review this as the modifications to metrics will appear as large diff blocks. But in the first part, we can move on through the below summary and examples and also the code changes. I tested this extensively, but it's possible that the combination of some CLI flags may raise errors, who knows. If merge this, we could do a release candidate first to let people test it.

- In terms of backward compatibility, I tried to be conservative. The handling of #141 is definitely a backward-incompatible fix but I think it's the correct behavior. Another incompatible change is how signatures are formatted on the terminal.

- Two things are hopefully handled correctly but actually untested on Windows: (1) Colored outputs (should be disabled on Windows through platform check), (2) multi-CPU significance test (should fall back to 1 CPU if on Windows)

Questions:

- Should we keep the single-system confidence (--confidence) functionality or is it confusing things as it does not actually provide very valuable information on itself.

- For #124 should we switch to chrF++ by default or continue computing the plain old chrF for backward compatibility?

Thanks!
 

General
-------

- Improve documentation and type annotations.
- Add README.md and CHANGELOG.md to `setup.py` so that they are shown on PyPI.
- Drop Python < 3.6 support and migrate to f-strings.
- Relax `portalocker` version pinning, add `regex, tabulate, numpy` dependencies.
- Drop input type manipulation through `isinstance` checks. If the user does not obey
  to the expected annotations, exceptions will be raised. Robustness attempts lead to
  confusions and obfuscated score errors in the past (#121)
- Variable # references per segment is supported for all metrics by default. It is
  still only available through the API.
- Use colored strings in tabular outputs (multi-system evaluation mode). The
  colored output is disabled if the platform is Windows or if the output is
  redirected into a file or `--no-color` is passed.

Tokenizers
-----------

- Add caching to tokenizers which seem to speed up things a bit in some cases.
- Add `regex` dependency and use it in the `V14International` tokenizer.
  Speed goes from ~4 seconds to ~0.6 seconds for a particular test set evaluation. (#46)

Metrics
--------

- General performance improvements for BLEU and CHRF (#46).
- Scale all metrics into the [0, 100] range (#140)
- BLEU: In case of no n-gram matches at all, skip smoothing and return 0.0 BLEU (#141).
- CHRF: Added multi-reference support, verified the scores against chrF++.py, added test case.
- CHRF: Added chrF+ support through `word_order` argument. Added test cases against chrF++.py.
  Exposed it through the CLI (--chrf-word-order). The default is still chrF and not chrF++ (#124)
- CHRF: Add possibility to disable effective order smoothing (pass --chrf-eps-smoothing). This way,
  the scores obtained are exactly the same as chrF++, Moses and NLTK implementations.
  I kept the effective ordering as the default since this only affects sentence-level
  scoring with very short sentences. (#144)
- TER: Move tokenizer signatures to the metric itself for consistency with other metrics.
- TER: Use `string.translate` for OP code mapping in one line of code.

Metric API
----------

- Use explicit argument names and defaults for the metrics instead of passing `argparse.Namespace`.
- Various sorts of refactoring in method names and arguments.
- A base abstract `Metric` class is introduced to guide further metric development
  This class defines the methods that should be implemented in the derived classes and
  offers boilerplate methods for the common functionality.
- All metrics now receive an optional `references` argument at initialization time
  to process and cache the references. Further evaluations of different systems against
  the same references becomes faster this way for example when using significance
  testing.

Signatures
----------
- The signature formatting changed (mostly to remove '+' separator as it was interfering with chrF++).
- The field separator is now '|' and key values are separated with ':' rather than '.'
- Boolean true / false values are shortened to yes / no, some other shortenings applied as well.
- Number of references is `var` if variable number of references is used.
- Add effective order (yes/no) to BLEU and chrF signatures.

CLI
----

- `--input/-i` can now ingest multiple systems. For this reason, the positional
  `references` should always preceed the `-i` flag:

```bash
# Correct
$ sacrebleu ref -i sys
# Incorrect (will fail)
$ sacrebleu -i sys ref
```

- Separate metric-specific arguments for clarity when `--help` is printed.
- Allow modifying TER arguments through CLI. We still keep the TERCOM defaults.
- Prefix metric-specific arguments with --chrf and --ter. To maintain CLI and historical compatibility,
  I did not add --bleu prefixes to BLEU arguments.
- When multiple metrics are asked, they are now aligned at `=` as follows:

```bash
 BLEU|case:mixed|nrefs:1|tok:13a|smooth:exp|version:2.0.0 = 23.2 55.4/28.6/17.1/10.6 (BP = 1.000 ratio = 1.059 hyp_len = 65345 ref_len = 61721)
chrF2|case:mixed|nrefs:1|nc:6|nw:0|space:no|version:2.0.0 = 52.6
```

- Added --format/-f flag. If single system is evaluated, -f json will print the results
  in a parseable JSON format. Arguments such as --short, --score-only are ignored and
  full information is dumped when `-f json` is given:

```bash
$ sacrebleu/sacrebleu.py ref -i sys
BLEU|nrefs:1|case:mixed|eff:no|tok:13a|smooth:exp|version:2.0.0 = 20.8 54.4/26.6/14.9/8.7 (BP = 1.000 ratio = 1.026 hyp_len = 62880 ref_len = 61287)

$ sacrebleu/sacrebleu.py ref -i sys -f json
{
 "name": "BLEU",
 "score": 20.8,
 "signature": "nrefs:1|case:mixed|eff:no|tok:13a|smooth:exp|version:2.0.0",
 "verbose_score": "54.4/26.6/14.9/8.7 (BP = 1.000 ratio = 1.026 hyp_len = 62880 ref_len = 61287)",
 "nrefs": "1",
 "case": "mixed",
 "eff": "no",
 "tok": "13a",
 "smooth": "exp",
 "version": "2.0.0"
}

$ sacrebleu/sacrebleu.py ref -i sys -f json | jq .score
20.8

# Multiple metrics and JSON
$ sacrebleu/sacrebleu.py ref -i sys -f json -m bleu chrf
[
{
 "name": "BLEU",
 "score": 20.8,
 "signature": "nrefs:1|case:mixed|eff:no|tok:13a|smooth:exp|version:2.0.0",
 "verbose_score": "54.4/26.6/14.9/8.7 (BP = 1.000 ratio = 1.026 hyp_len = 62880 ref_len = 61287)",
 "nrefs": "1",
 "case": "mixed",
 "eff": "no",
 "tok": "13a",
 "smooth": "exp",
 "version": "2.0.0"
},
{
 "name": "chrF2",
 "score": 52.0,
 "signature": "nrefs:1|case:mixed|eff:yes|nc:6|nw:0|space:no|version:2.0.0",
 "nrefs": "1",
 "case": "mixed",
 "eff": "yes",
 "nc": "6",
 "nw": "0",
 "space": "no",
 "version": "2.0.0"
}
]

# Iterate over the list and fetch the score key
$ sacrebleu/sacrebleu.py ref -i sys -f json -m bleu chrf | jq '.[] | .score'
20.8
52
```

- For multi-system mode, `json` falls back to plain text. Other options exist
  in this mode such as `latex, rst, html`. (More on this later)

Multi-system evaluation mode
----------------------------

- sacreBLEU now supports evaluating multiple systems for a given test set in an efficient way.
  Through the use of `tabulate` package, the results are nicely rendered into a table
  in plain text, LaTeX, HTML or RST (cf. --format/-f argument).

- The systems can be either given as a list of plain text files to `-i/--input` or
  as a tab-separated single stream redirected into `STDIN`. In the former case,
  the basenames of the files will be automatically used as system names.

- If you give the same file twice, sacreBLEU will issue an error.

**Explicit filenames**:

```
$ sacrebleu/sacrebleu.py -t wmt17 -l cs-en -i newstest2017.* -m bleu chrf
sacreBLEU: Found 4 systems.
+-----------------------------------+--------+---------+
|                            System |  BLEU  |  chrF2  |
+===================================+========+=========+
|     newstest2017.online-A.0.cs-en |  25.1  |  53.4   |
+-----------------------------------+--------+---------+
|     newstest2017.online-B.0.cs-en |  27.4  |  54.5   |
+-----------------------------------+--------+---------+
|     newstest2017.PJATK.4760.cs-en |  23.2  |  52.6   |
+-----------------------------------+--------+---------+
| newstest2017.uedin-nmt.4955.cs-en |  30.9  |  56.8   |
+-----------------------------------+--------+---------+

-----------------
Metric signatures
-----------------
 - BLEU       nrefs:1|case:mixed|eff:no|tok:13a|smooth:exp|version:2.0.0
 - chrF2      nrefs:1|case:mixed|eff:yes|nc:6|nw:0|space:no|version:2.0.0
```

**Tab-separated STDIN**:

```
$ paste newstest2017.* | sacrebleu/sacrebleu.py -t wmt17 -l cs-en -m bleu chrf
sacreBLEU: Found 4 systems.
+----------+--------+---------+
|   System |  BLEU  |  chrF2  |
+==========+========+=========+
| System 1 |  25.1  |  53.4   |
+----------+--------+---------+
| System 2 |  27.4  |  54.5   |
+----------+--------+---------+
| System 3 |  23.2  |  52.6   |
+----------+--------+---------+
| System 4 |  30.9  |  56.8   |
+----------+--------+---------+
```

**LaTeX mode**:

```
$ paste newstest2017.* | sacrebleu/sacrebleu.py -t wmt17 -l cs-en -m bleu chrf -f latex
sacreBLEU: Found 4 systems.
\begin{tabular}{rcc}
\toprule
   System &  BLEU  &  chrF2  \\
\midrule
 System 1 &  25.1  &  53.4   \\
 System 2 &  27.4  &  54.5   \\
 System 3 &  23.2  &  52.6   \\
 System 4 &  30.9  &  56.8   \\
\bottomrule
\end{tabular}
```

Single-system bootstrap confidence intervals (requires numpy) (#40 and #78)
---------------------------------------------------------------------------
- 95% confidence intervals are provided only for the single-system evaluation mode.
  If you have multiple systems, we recommend using paired tests that will provide
  both the confidence intervals and the p-values.

- The feature is enabled by passing `--confidence` to the CLI. The default number
  of bootstrap resamples is 2000. This can be changed with the `--confidence-n` flag.

- The random number generator's seed is by default fixed to 12345. The seed
  can be modified by exporting `SACREBLEU_SEED` environment variable.
  If the exported value is `[Nn]one`, the seed is uninitialized, yielding
  non-deterministic results.

- Unit tests are added to compare the results to Moses' significance Perl script.

**Fixed seed**:
```bash
$ sacrebleu/sacrebleu.py ref -i fbk --confidence -m bleu chrf
  BLEU|nrefs:1|bs:2000|seed:12345|case:mixed|eff:no|tok:13a|smooth:exp|version:2.0.0 = 26.3 (μ = 26.33 ± 0.65) 59.3/32.3/19.9/12.6 (BP = 1.000 ratio = 1.003 hyp_len = 61450 ref_len = 61287)
chrF2|nrefs:1|bs:2000|seed:12345|case:mixed|eff:yes|nc:6|nw:0|space:no|version:2.0.0 = 54.7 (μ = 54.72 ± 0.48)
```

**Random seed**:
```bash
$ SACREBLEU_SEED=None sacrebleu/sacrebleu.py ref -i fbk --confidence -m bleu chrf ter
        BLEU|nrefs:1|bs:2000|seed:none|case:mixed|eff:no|tok:13a|smooth:exp|version:2.0.0 = 26.3 (μ = 26.30 ± 0.65) 59.3/32.3/19.9/12.6 (BP = 1.000 ratio = 1.003 hyp_len = 61450 ref_len = 61287)
      chrF2|nrefs:1|bs:2000|seed:none|case:mixed|eff:yes|nc:6|nw:0|space:no|version:2.0.0 = 54.7 (μ = 54.72 ± 0.50)
TER|nrefs:1|bs:2000|seed:none|case:lc|tok:tercom|norm:no|punct:yes|asian:no|version:2.0.0 = 63.6 (μ = 63.63 ± 0.78)
```

Multi-system paired significance tests (#40 and #78)
----------------------------------------------------
- When you have multiple systems to evaluate for a given test set and language pair,
  you can now use paired significance tests to obtain p-values.

- The **first** system provided to `--input/-i` (or the first column hypotheses
  if pasted `STDIN` method is used) will be flagged as the **baseline system**.
  When using `--input/-i`, sacreBLEU will automatically discard the baseline system
  if it appears more than one time. This is useful when using shell globs.

- Two types of paired tests are provided: Bootstrap resampling (`bs`) and
  approximate randomization (`ar`). `bs` replicates the behavior of Moses'
  significance Perl script whereas `ar` follows the [Multeval](https://github.com/mjclark/multeval)
  for performing approximate randomization. The feature is enabled by passing
  one of these two methods to the `--paired` flag.

- The default number of samples/trials for `bs` and `ar` are 2,000 and 10,000, respectively.
  This can be changed by using the `--paired-n/-pan` flag.

- The `bs` test will also print 95% CI around the true mean as additional information.
  To enable same type of CI's for the AR test, pass `--paired-ar-confidence-n 0`
  for example, to use the default value of 2000 resamples.

- Verbose information printed during the tests can be disabled by `--quiet`.

**Example of evaluating 16 WMT17 submissions with 2 metrics**:

```bash
# the LIUM system will not be counted as a candidate system as it was given
# as the baseline
$ sacrebleu -t wmt17 -l en-de -i newstest2017.LIUM-NMT.4900.en-de newstest2017.* --paired bs -m bleu chrf
(Verbose messages suppressed)

+--------------------------------------------+-----------------------+-----------------------+
|                                     System |  BLEU / μ / ± 95% CI  | chrF2 / μ / ± 95% CI  |
+============================================+=======================+=======================+
| Baseline: newstest2017.LIUM-NMT.4900.en-de |  26.6 / 26.6 / 0.65   |  55.9 / 55.9 / 0.47   |
+--------------------------------------------+-----------------------+-----------------------+
|              newstest2017.C-3MA.4959.en-de |  22.7 / 22.7 / 0.61   |  52.0 / 52.0 / 0.46   |
|                                            |     (p = 0.0005)*     |     (p = 0.0005)*     |
+--------------------------------------------+-----------------------+-----------------------+
|                newstest2017.FBK.4870.en-de |  26.3 / 26.3 / 0.65   |  54.7 / 54.7 / 0.48   |
|                                            |     (p = 0.0945)      |     (p = 0.0005)*     |
+--------------------------------------------+-----------------------+-----------------------+
|                newstest2017.KIT.4950.en-de |  26.1 / 26.1 / 0.66   |  55.8 / 55.8 / 0.46   |
|                                            |     (p = 0.0105)*     |     (p = 0.1089)      |
+--------------------------------------------+-----------------------+-----------------------+
|   newstest2017.LMU-nmt-reranked.4934.en-de |  27.1 / 27.1 / 0.65   |  56.4 / 56.4 / 0.46   |
|                                            |     (p = 0.0070)*     |     (p = 0.0015)*     |
+--------------------------------------------+-----------------------+-----------------------+
|     newstest2017.LMU-nmt-single.4893.en-de |  26.6 / 26.5 / 0.66   |  55.9 / 55.9 / 0.44   |
|                                            |     (p = 0.3353)      |     (p = 0.4098)      |
+--------------------------------------------+-----------------------+-----------------------+
|              newstest2017.online-A.0.en-de |  20.8 / 20.8 / 0.59   |  52.0 / 52.0 / 0.43   |
|                                            |     (p = 0.0005)*     |     (p = 0.0005)*     |
+--------------------------------------------+-----------------------+-----------------------+
|              newstest2017.online-B.0.en-de |  26.7 / 26.7 / 0.67   |  56.3 / 56.3 / 0.45   |
|                                            |     (p = 0.3073)      |     (p = 0.0240)*     |
+--------------------------------------------+-----------------------+-----------------------+
|              newstest2017.online-F.0.en-de |  15.5 / 15.5 / 0.49   |  49.3 / 49.3 / 0.39   |
|                                            |     (p = 0.0005)*     |     (p = 0.0005)*     |
+--------------------------------------------+-----------------------+-----------------------+
|              newstest2017.online-G.0.en-de |  18.2 / 18.2 / 0.54   |  51.6 / 51.6 / 0.40   |
|                                            |     (p = 0.0005)*     |     (p = 0.0005)*     |
+--------------------------------------------+-----------------------+-----------------------+
|   newstest2017.PROMT-Rule-based.4735.en-de |  16.6 / 16.6 / 0.51   |  50.4 / 50.4 / 0.40   |
|                                            |     (p = 0.0005)*     |     (p = 0.0005)*     |
+--------------------------------------------+-----------------------+-----------------------+
|  newstest2017.RWTH-nmt-ensemble.4921.en-de |  26.0 / 26.0 / 0.66   |  55.6 / 55.6 / 0.45   |
|                                            |     (p = 0.0050)*     |     (p = 0.0120)*     |
+--------------------------------------------+-----------------------+-----------------------+
|            newstest2017.SYSTRAN.4847.en-de |  26.7 / 26.7 / 0.66   |  55.6 / 55.6 / 0.45   |
|                                            |     (p = 0.2144)      |     (p = 0.0085)*     |
+--------------------------------------------+-----------------------+-----------------------+
|           newstest2017.TALP-UPC.4834.en-de |  21.2 / 21.2 / 0.58   |  51.7 / 51.7 / 0.43   |
|                                            |     (p = 0.0005)*     |     (p = 0.0005)*     |
+--------------------------------------------+-----------------------+-----------------------+
|          newstest2017.uedin-nmt.4722.en-de |  28.3 / 28.3 / 0.70   |  57.7 / 57.7 / 0.46   |
|                                            |     (p = 0.0005)*     |     (p = 0.0005)*     |
+--------------------------------------------+-----------------------+-----------------------+
|                newstest2017.xmu.4910.en-de |  26.7 / 26.7 / 0.65   |  56.0 / 56.0 / 0.45   |
|                                            |     (p = 0.2409)      |     (p = 0.3058)      |
+--------------------------------------------+-----------------------+-----------------------+

------------------------------------------------------------
Paired bootstrap resampling test with 2000 resampling trials
------------------------------------------------------------
 - Each system is pairwise compared to Baseline: newstest2017.LIUM-NMT.4900.en-de.
   Actual system score / estimated true mean / 95% CI are provided for each metric.

 - Null hypothesis: the system and the baseline translations are essentially
   generated by the same underlying process. The p-value is roughly the probability
   of the absolute score difference (delta) between a system and the {bline} occurring due to chance.

 - Assuming a significance threshold of 0.05, the Null hypothesis can be rejected
   for p-values < 0.05 (marked with "*"). This means that the delta is unlikely to be attributed
   to chance, hence the system is significantly "different" than the baseline.
   Otherwise, the p-values are highlighted in red (if the terminal supports colors).

 - NOTE: Significance does not tell whether a system is "better" than the baseline but rather
   emphasizes the "difference" of the systems in terms of the replicability of the delta.

-----------------
Metric signatures
-----------------
 - BLEU       nrefs:1|bs:2000|seed:12345|case:mixed|eff:no|tok:13a|smooth:exp|version:2.0.0
 - chrF2      nrefs:1|bs:2000|seed:12345|case:mixed|eff:yes|nc:6|nw:0|space:no|version:2.0.0
```

- The above command takes ~45 seconds to run which is quite acceptable. However,
  if you also enable TER, it takes ~3.5 minutes to complete. Therefore, it is also
  possible to run the tests using multiple workers (only on Linux and Mac OS X).
  Passing `0` to `--paired-jobs/-paj` will launch as many workers as the number
  of systems (up to the limit of the number of CPUs on the machine) whereas
  passing a value `> 0` will manually set the number of workers in the pool.
  For the above example, it takes ~40 seconds to complete using 15 workers
  (for 15 candidate systems, excluding the baseline).

**Approximate randomization method**:

```bash
# No CI estimation, just p-value's
$ sacrebleu -t wmt17 -l en-de -i newstest2017.LIUM-NMT.4900.en-de newstest2017.* --paired ar -m bleu chrf

# With CI estimation using the default of 2000 bootstrap resamples
$ sacrebleu -t wmt17 -l en-de -i newstest2017.LIUM-NMT.4900.en-de newstest2017.* --paired ar --paired-ar-confidence-n 2000 -m bleu chrf
```

SacreBLEU 2.0.0 performance tests
---------------------------------
- The tests are performed across 4 different systems using the API. This is
  where caching kicks in. (The reason why no-cache also speeds up things
  after the 1st evaluation is the caching in the tokenizers.)
- The 1st column between 1.5.1 and no-cache should be seen as the vanilla
  performance diff between versions.

```
BLEU {}
 >    [1.5.1] 0.890 0.890 0.890 0.891  || mean: 0.890 -- median: 0.890 -- stdev: 0.000
 > [no-cache] 0.612 0.458 0.452 0.435  || mean: 0.489 -- median: 0.458 -- stdev: 0.071
 >   [cached] 0.604 0.315 0.310 0.303  || mean: 0.383 -- median: 0.315 -- stdev: 0.128
BLEU {'tokenize': 'intl'}
 >    [1.5.1] 4.048 4.039 4.039 4.072  || mean: 4.108 -- median: 4.048 -- stdev: 0.119
 > [no-cache] 0.516 0.426 0.424 0.389  || mean: 0.439 -- median: 0.426 -- stdev: 0.047
 >   [cached] 0.504 0.281 0.271 0.302  || mean: 0.339 -- median: 0.302 -- stdev: 0.095
BLEU {'tokenize': 'none', 'force': True}
 >    [1.5.1] 0.516 0.518 0.523 0.514  || mean: 0.516 -- median: 0.516 -- stdev: 0.004
 > [no-cache] 0.264 0.268 0.293 0.265  || mean: 0.273 -- median: 0.268 -- stdev: 0.012
 >   [cached] 0.257 0.158 0.153 0.152  || mean: 0.180 -- median: 0.158 -- stdev: 0.045
CHRF {}
 >    [1.5.1] 1.105 1.102 1.099 1.100  || mean: 1.101 -- median: 1.100 -- stdev: 0.002
 > [no-cache] 1.111 1.086 1.078 1.092  || mean: 1.092 -- median: 1.092 -- stdev: 0.012
 >   [cached] 1.050 0.660 0.659 0.644  || mean: 0.753 -- median: 0.660 -- stdev: 0.171
CHRF {'whitespace': True}
 >    [1.5.1] 1.221 1.222 1.223 1.218  || mean: 1.225 -- median: 1.222 -- stdev: 0.009
 > [no-cache] 1.250 1.280 1.250 1.234  || mean: 1.254 -- median: 1.250 -- stdev: 0.016
 >   [cached] 1.251 0.778 0.772 0.756  || mean: 0.889 -- median: 0.778 -- stdev: 0.209
```


martinpopel (commented):
Just a partial review. I will continue later if time allows.

sacrebleu/color.py
```diff
@@ -0,0 +1,36 @@
+import os
+import sys
+
+
+_ENABLE = True
+if not sys.stdout.isatty() or os.name == 'nt':
```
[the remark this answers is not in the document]
ozancaglayan: I actually tried it first but couldn't understand fully its benefits and decided to use this simple class instead of adding one more dependency just for colors. But if that works on Windows as well, we can switch to it if everybody is OK.

sacrebleu/metrics/bleu.py
```diff
@@ -1,84 +1,111 @@
...
-    """A convenience class to represent BLEU scores (without signature)."""
-    def __init__(self, score, counts, totals, precisions, bp, sys_len, ref_len):
-        super().__init__(score)
-
-        self.prefix = 'BLEU'
+class BLEUScore(Score):
+    """A convenience class to represent BLEU scores."""
+    def __init__(self, score: float, counts: List[int], totals: List[int],
+                 precisions: List[float], bp: float,
+                 sys_len: float, ref_len: float):
```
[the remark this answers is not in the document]
ozancaglayan: Good catch. This is a leftover from an earlier state of the PR. Fixed now.

sacrebleu/metrics/__init__.py
```diff
@@ -1,11 +1,9 @@
-# -*- coding: utf-8 -*-
-
-from .bleu import BLEU, BLEUScore
-from .chrf import CHRF, CHRFScore
-from .ter import TER, TERScore
+from .bleu import BLEU, BLEUScore   # noqa: F401
```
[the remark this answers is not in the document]
ozancaglayan: I don't know whether it will fix the unused import warning?

sacrebleu/color.py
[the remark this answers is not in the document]
ozancaglayan: I now tried on a virtualbox Win10 VM and unfortunately colorama does not seem to work there out of the box. They say that you also require the termcolor package.

sacrebleu/metrics/__init__.py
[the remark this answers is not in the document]
martinpopel: Oh, I thought F401 is because of the relative import.

sacrebleu/metrics/bleu.py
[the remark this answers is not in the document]
martinpopel: Where is it fixed now?
When fixing, change also `self.sys_len = int(sys_len)` to `self.sys_len = sys_len` (and similarly for `ref_len`).

sacrebleu/color.py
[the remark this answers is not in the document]
martinpopel: The `termcolor` package is not needed, it just gives you syntactic sugar and more color codes.
I don't have access to a Windows box, but it seems
`python -c 'import colorama;print("\033[31m some red text")'`
[just works](https://travis-ci.org/github/udapi/udapi-python/jobs/765305822#L124).
BTW: Travis-CI newly supports Windows, so we could try it after handling [test.sh](https://github.com/mjpost/sacrebleu/blob/master/test.sh) (we could omit it from the tests or try if it works with git-bash or rewrite it to Python).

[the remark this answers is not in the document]
ozancaglayan: Ok got it. `colorama` should be imported before `sys`. And also, in `sacrebleu.py` we need to remove re-opening `sys.stdin` and `sys.stdout` (potentially only the latter), otherwise it does not work. I don't know if it's still necessary to re-open standard systems that way, I've never used that in my projects. It's probably dating back from Python 2.x?

sacrebleu/metrics/bleu.py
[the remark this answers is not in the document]
ozancaglayan: ok fixed now.

sacrebleu/color.py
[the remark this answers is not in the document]
ozancaglayan: Figured out how to run things smoothly on Windows and OS X with travis, I'll try to push those as well to this branch.

README.md
```diff
@@ -1,115 +1,624 @@
...
+
+Let's now remove the first reference sentence for the first system sentence `The dog bit the man.` by replacing it with either `None` or the empty string `''`.
+This allows using a variable number of reference segments per hypothesis. Observe how the signature changes from `nrefs:2` to `nrefs:var`:
+
+```python
+In [1]: from sacrebleu.metrics import BLEU, CHRF, TER
+   ...: 
+   ...: refs = [ # First set of references
+                 # 1st sentence does not have a ref here
+   ...:          ['', 'It was not unexpected.', 'The man bit him first.'],
```
martinpopel, line 582: So the first sentence has no reference at all?
Such example fails with RuntimeError: Empty or `None` reference sentence found.
I think the use case we want to show here is that the first sentences has only a single reference (and is missing the second reference), i.e.
`['The dog bit the man.', 'It was not unexpected.', 'The man bit him first.'],`

README.md
```diff
@@ -1,115 +1,624 @@
...
+# Command-line Usage
 
-This installs a shell script, `sacrebleu`.
-(You can also run `python3 -m sacrebleu`, so long as this root directory is in your `$PYTHONPATH`).
+## Basics
 
 Get a list of available test sets:
 
-    sacrebleu --list
+```
```
martinpopel, line 59: It is great that all the (old and new) features of SacreBLEU are now well documented.
However, I think the README is now a bit too long and newcomers may get impression that SacreBLEU is too difficult to use.
An alternative would be to break the current content into several pages which will be just linked from the main README.
The README should show perhaps just:
```
$ pip install sacrebleu
$ cat output.detok.txt | sacrebleu -t wmt17 -l en-de
BLEU|nrefs:1|case:mixed|eff:no|tok:13a|smooth:exp|version:2.0.0 = 20.8 54.4/26.6/14.9/8.7 (BP = 1.000 ratio = 1.026 hyp_len = 62880 ref_len = 61287)
```
The detailed documentation of all features could be also maintained as GitHub wiki (as done e.g. in [VW](https://github.com/VowpalWabbit/vowpal_wabbit/wiki)) or ReadTheDocs.

That needs some effort, so maybe it can be done in another PR after this PR is merged.

Within this PR, I would just suggest to delete the output of `sacrebleu --list`.
It will soon become outdated. If we want provide the listing in Markdown,
we could generate it automatically (within deployment) to a separate file.

README.md
```diff
@@ -1,115 +1,624 @@
...
+# Use the ja-mecab tokenizer
+$ sacrebleu kyoto-test.ref.ja -i kyoto-test.hyp.ja --tokenize ja-mecab -b
+14.5
+
+# Alternatively, if you provide the language-pair, sacreBLEU will use ja-mecab automatically
+$ sacrebleu kyoto-test.ref.ja -i kyoto-test.hyp.ja -l en-ja -b
+14.5
+```
+
+### chrF / chrF+
```
martinpopel, line 292: or chrF++?

sacrebleu/tokenizers/tokenizer_13a.py
```diff
@@ -1,17 +1,17 @@
...
-class Tokenizer13a(NoneTokenizer):
+class Tokenizer13a(BaseTokenizer):
 
     def signature(self):
         return '13a'
 
     def __init__(self):
         self._post_tokenizer = TokenizerRegexp()
 
+    @lru_cache(maxsize=None)
```
martinpopel, line 14: What is the use case (exact command) when lru_cache should help?

My benchmarks (wmt17 de-en) give me slightly faster times (and less memory) when excluding this lru_cache
here and also in tokenizer_re and tokenizer_intl.
(My expectation was that the cache in tokenizer_re is extra, if it is called only from another function which is already cached.)
Interestingly, this holds also for `--paired bs` and `-i newstest2017/de-en/*` with 11 systems, so there is a chance that some sentences will be the same across different systems and the cache will be used.

sacrebleu/tokenizers/tokenizer_intl.py
```diff
@@ -1,76 +1,50 @@
...
+            # Separate out symbols
+            (regex.compile(r'(\p{S})'), r' \1 '),
+        ]
+
+    @lru_cache(maxsize=None)
+    def __call__(self, line: str) -> str:
+        for (_re, repl) in self._re:
+            line = _re.sub(repl, line)
+
+        return ' '.join(line.strip().split())
```
martinpopel, line 50: return ' '.join(line.split())

README.md
ozancaglayan: The naming of chrF+* is a bit weird: Their idea was, for every word n-gram added, you add a `+`. So chrF++ is when `--chrf-word-order == 2`. That's why I decided to refer to the revised version of the metric using `chrF+` in a generic way...

sacrebleu/tokenizers/tokenizer_13a.py
ozancaglayan: My mind is confused on this as well. I initially added this as a quick hack so that references are not re-tokenized over and over when using multi-system mode. But then I realized that this wasn't the good way to do it and I switched to reference caching. But weirdly, if I remove the caches, I slightly have slower runtimes. I thought that the reason was what you wrote i.e. some hypotheses are the same across systems. But you're also saying that removing the caches are faster for you? Sorry, didn't understand your case, could you clarify?

Also, how do you measure the time? I used `scripts/perf_test.py` once on the branch and once after manually commenting all lru_cache's.

README.md
ozancaglayan: No first sentence has a reference but coming from the second set of references. The order shouldn't matter, I don't know why I removed one of the refs from the first list instead of the second. I'll see this.

README.md
ozancaglayan: exactly, I didn't have the energy to do the auto generated docs bits for this PR but that should arrive sooner or later.

For the `--list`, since we are doing the PRs and merges, I thought we could ask people to just add a line to that portion in the README whenever they add a new dataset. Isn't that reasonable? It's not that we add a new dataset every now and then.

sacrebleu/tokenizers/tokenizer_intl.py
ozancaglayan: done

README.md
martinpopel: OK. Just that chrF++ is mentioned several times above, so the readers/users may get confused.
I would go with
`--chrf-word-order=0` ... chrF
`--chrf-word-order=2` ... chrF++
these are the two common options used in papers. We don't need to introduce a name for a chrF with any other word order.

Thus I suggest also:

> If you enable chrF+ mode by passing `--chrf-word-order` (The number of `+` letters denotes the word n-gram order that you select),
> a very simple punctuation tokenization will be internally applied.

If you enable non-zero `--chrf-word-order`, a very simple punctuation tokenization will be internally applied.

But I don't have any strong opinion on this. Feel free to ignore this suggestion.

ozancaglayan: Makes sense. I'll revise.

sacrebleu/tokenizers/tokenizer_13a.py
martinpopel: Interesting: `scripts/perf_test.py` uses `newstest2017/cs-en`, which caching makes a bit faster. I used `newstest2017/de-en`, which caching makes a bit slower.
I tried python `timeit` and also extrinsic (running it several times and computing mean/median):
`/usr/bin/time -f "\t%E real,\t%M mem" sacrebleu/sacrebleu.py -t wmt17 -l de-en -i data/wmt17-submitted-data/txt/system-outputs/newstest2017/de-en/*`.

The differences are not big, but caching could help with more system outputs (or test sets with many repeated sentences).  The memory is also not much bigger, so we can keep the cache there.

README.md
martinpopel: Currently, it is missing in both sets of references.
I agree the order of alternative references shouldn't matter, so I am OK with showing a missing sentence in the first set of references.

ozancaglayan: ahh, sorry that's indeed an error.

README.md
martinpopel: I think its better to ask users to run `sacrebleu --list` themselves and maybe remind them to run `pip install --upgrade sacrebleu` first.
That said, you are right it shouldn't be much work to keep the listing updated (although I would still prefer to have it at a separate page).

sacrebleu/utils.py
```diff
@@ -27,6 +26,194 @@
...
+    return new_scores
+
+
+def print_results_table(results: dict, signatures: dict, args: Namespace):
+    """Prints out a nicely formatted table for multi-system evaluation mode."""
+
+    tablefmt = args.format
+    if tablefmt in ('text', 'json'):
+        # Fallback to simple table if json is given
+        tablefmt = 'grid'
```
martinpopel, line 96: What about `fancy_grid`?
We are already using unicode (μ / ±).

sacrebleu/utils.py
```diff
@@ -27,6 +26,194 @@
...
+        print('-' * len(msg) + '\n' + msg + '\n' + '-' * len(msg))
+        print(f' - Each system is pairwise compared to {bline_name}.')
+        if args.paired == 'bs' or args.paired_ar_confidence_n != -1:
+            print('   Actual system score / estimated true mean / 95% CI are provided for each metric.')
+        else:
+            print('   Actual system score is provided for each metric.')
+        print()
+        print(f' - {null_hyp}: the system and the {bline} translations are essentially')
+        print('   generated by the same underlying process. The p-value is roughly the probability')
+        print('   of the absolute score difference (delta) between a system and the {bline} occurring due to chance.')
```
martinpopel, line 149: Strictly speaking "of the absolute score difference (delta) *or higher*" and "given the null hypothesis",
but the explanatory comment is long enough so this suggestion can be ignored.

ozancaglayan: I'm willing to make this better since I basically looked into different sources to come up with this.
So that would be:
` - Given the null hyp: ....` for the second suggestion right?

sacrebleu/utils.py
ozancaglayan: It was `fancy_grid` actually but after adding `json`, I probably falled back to `grid` for some reason. Thanks.

sacrebleu/utils.py
martinpopel: Yes, p-value is the probability (of...) under the assumption that the null hypothesis is correct

README.md
ozancaglayan: fixed.

martinpopel:
I am not sure if the "BLEU / μ / ± 95% CI" format (e.g. "33.2 / 33.2 / 0.70") is ideal.
* For better readability, it could be "33.2 (33.2 ± 0.70)".
* I think the plus-minus diff should have the same precision as the main score, i.e. "33.2 (33.2 ± 0.7)". Ideally, we should compute the number of significant figures, but it seems easier to just follow the precision specified with `-w`. E.g. with `-w 4`, "33.1820  ± 0.70" looks weird.
* I know the sample mean does not need to be always the same as the original BLEU, but reporting just "33.2 ± 0.7" seems much nicer. I don't see much sense in reporting the mean.

The current implementation shows just paired tests against the baseline. This is perhaps enough for most users, so I look forward to see this PR merged.
However, in future it would be nice to:
* Sort all the systems by BLEU.
* Draw a double/bold line between each pair of systems (on subsequent lines) with significantly different BLEU.
* Add a rank range for each system, e.g. 1-5 means a given systems is not significantly different than systems on lines 1 and 5 (and virtually always also 2,3,4).
* This means computing the paired test between each pair of systems, but I hope this could be done quite efficiently with numpy.
* Obviously, no rank range can cross the bold lines, but there can be different rank ranges within one "cluster" (separated by bold lines), e.g. in the example below, A is significantly better than C, so if B is excluded from the evaluation, there would be a bold line between A and C. BTW: rank ranges used to be [reported for WMT](http://www.statmt.org/wmt16/pdf/W16-2301.pdf#page=11) human evaluation before switching to DA, although there is no reason for not reporting rank ranges for DA (or BLEU or chrF etc).
```
----------------------------
1-2 | systemA | 33.2 ± 0.7 |
1-3 | systemB | 33.0 ± 0.8 |
2-3 | systemC | 32.6 ± 0.7 |
----------------------------
4   | systemD | 30.5 ± 0.7 |
----------------------------
```
* There could be an option to turn this behavior on (vs. reporting the systems in the order provided by the user). There is a related question whether to report p-values for each system relative to the baseline or relative to the system on the previous line (so whenever p<0.05, there will be a bold line above such system).

mjpost:
Hi, this looks incredible! I haven't been able to follow this in depth yet, but have one request. Could we:

* Have the signature output in JSON by default
* Have the score embedded in the JSON obey the `-w` flag?

I think this is more readable, both for humans (opinion) and machine (fact, via `jq`, as you note). I like the new delimiter for the text-based string.

I was going to suggest that the JSON output be single line, but I have just learned that `jq` can easily parse multi-line output. Very nice!

ozancaglayan:
@martinpopel I've fixed all your concerns except this:

> I know the sample mean does not need to be always the same as the original BLEU, but reporting just "33.2 ± 0.7" seems much nicer. I don't see much sense in reporting the mean.

So do you want to report `actual_score +- ci` instead of `actual_score (mean +- ci)` ? Yes with large numbers of bootstrap resamples, mean ~= actual_score often, but avoiding mean may risk of being unsound in terms of statistics?

@mjpost Thanks! A post-ACL spare time procrastination :) 

> Have the signature output in JSON by default

This I didn't understand. The signature is there in both exploded form as keys values, and also as a single string form under the key `signature`, if `--format json` is given. Are you suggesting something else?

> Have the score embedded in the JSON obey the -w flag?

This is fixed now.

mjpost:
Hi Ozan—Sorry, I was sloppy in my wording. What I meant was, have the JSON *format* be the default output format. I think it’s much nicer and more readable, and it’s the kind of thing we can only change, really, with a major version bump.

My dislike of the text-based format as the default is (a) it’s not that human readable and (b) it’s not machine-readable.

ozancaglayan:
Oh I see, that's a big one then =) I did not test Json output thoroughly and I am not sure if that radical change will make people, who probably already have a lot of codes that just parsed the output of sacreBLEU, happy. Even the signature format changed, the output should still be parseable with simple UNIX tools. On the other hand, I don't know for example, how do I pick `chrF` let's say from the Json output using `jq` if `-m bleu chrf` is given.

If you like, we can read `~/.sacrebleu.conf` to permanently change some settings such as having `preserve_text_output = True` for people who like the old way of doing things.

If we are willing to break things a bit more aggressively, I can then change the following compatibility hack as well?

> Prefix metric-specific arguments with --chrf and --ter. To maintain CLI compatibility,
I did not add --bleu prefixes to BLEU arguments but I manually modify them in sacrebleu.py.
We can drop this hack and expect from users to explicitly pass --bleu-tokenize --bleu-smooth-method etc.

ozancaglayan:
Hello!

How do we proceed from here? I had some questions above for both @martinpopel and @mjpost .

martinpopel:
> with large numbers of bootstrap resamples, mean ~= actual_score

Yes. If there is a notable difference between the bootstrap mean score and actual_score, it clearly shows we need more resamples (and maybe also a bigger test set if possible). I think it can be proven that with a growing number of resamples the mean score goes to the actual_value (which is the "correct" mean, in other words).

> often, but avoiding mean may risk of being unsound in terms of statistics?

Maybe. I admit we would loose the indication that the number of resamples was too low. But there are many other aspects where we simplify in terms of statistics (for example see footnote 2 in [Berg-Kirkpatrick et al., 2012](https://www.aclweb.org/anthology/D12-1091.pdf). So I would still prefer reporting just `actual_score +- ci` instead of `actual_score (mean +- ci)`, but I don't insist on this. I am OK with both.

> Also, for the other significance functionality, I think that may be implemented as a standalone script,

OK

I agree we should release v2.0.0 very soon and possibly solve minor issues in later versions.
(There are PR waiting for v2 to be merged, e.g. #153.)

ozancaglayan:
@mjpost Do you have some time for resolving some minor discussion points (such as changing the output to JSON by default) and possibly an approval for the PR?

ozancaglayan:
Leftover points:

- [ ] @mjpost would like to change the output format to JSON by default. What is your take on this @martinpopel ?
- [ ] If we break things such as the above one, as this is a major version bump, I would also revisit the cmdline params for BLEU which do not begin with `--bleu`  since sacreBLEU only included BLEU up to some point, and add explicit `--bleu` prefixes to all of them similar to `--chrf-*` and `--ter-*`. 
- [ ] (Minor) @martinpopel slighly leans toward giving the `actual_score` and not the sample mean i.e. quoting: `I know the sample mean does not need to be always the same as the original BLEU, but reporting just "33.2 ± 0.7" seems much nicer. I don't see much sense in reporting the mean.`

martinpopel:
I cannot decide what should be the default format, but as long as the users can override it with `-f` (to JSON or to the original one-line), I am OK with that. The `sacrebleu ... | jq .score` trick is nice and could be mentioned in the documentation.

mjpost:
I have had a hard time making time to process all this. What if we met over video for an hour and resolved everything? It would be nice to have this wrapped up. I'll email you two.

One thing I am worried about is feature creep. This tool's original goal was (a) a canonical bleu tool that (b) automatically downloaded references and (c) was super-easy to use. It's nice that we've added other metrics, but I wonder if it was a mistake, since there is really no end do it. This has bearing on the flag names: it makes sense to me to *not* add `--bleu` as a prefix, since BLEU is the default and main purpose of the tool.

I do think it makes sense to add features like (paired) bootstrap sampling and others. But ideally, sacrebleu will soon be just in maintenance mode, with occasional updates to add new references.

sacrebleu/utils.py
ozancaglayan: fixed

ozancaglayan:
I'm a bit busy these days, I'll work on this when I find some time the following days/week.

mjpost:
Sounds good, looking forward to it!

martinpopel (approved):

ozancaglayan:
Thanks @martinpopel but FYI, I am not done yet :)

ozancaglayan:
I've fixed some final issues and reworked the paired significance related arguments to make them a bit more readable and simpler. Tomorrow, I'll update the README and probably it'll be ready to merge.

Unfortunately Travis CI seems to be down for some time, it is not triggered at all. Any idea on that?

```
Paired significance testing for multi-system evaluation:
  --paired-ar, -par     Perform paired test using approximate randomization (AR). This option is mutually exclusive with --paired-bs (Default: False).
  --paired-bs, -pbs     Perform paired test using bootstrap resampling. This option is mutually exclusive with --paired-ar (Default: False).
  --paired-ar-n PAIRED_AR_N, -part PAIRED_AR_N
                        Number of trials for approximate randomization test (Default: 10000).
  --paired-ar-ci, -parci
                        If given, computes confidence interval for each system using bootstrap resampling in addition to the approximate randomization. The number of resamples is set to
                        2000.
  --paired-bs-n PAIRED_BS_N, -pbsn PAIRED_BS_N
                        Number of bootstrap resamples for paired bootstrap resampling test (Default: 2000).
  --paired-jobs PAIRED_JOBS, -j PAIRED_JOBS
                        If 0, launches as many workers as the number of systems. If > 0, sets the number of workers manually. This feature is currently not supported on Windows.
```

ozancaglayan:
My one remaining concern was pointed by @martinpopel in our meeting: When doing the paired AR test, the significance is *not* computed through bootstrap resampling but if the user provides `--paired-ar-ci` flag, the table will also include the confidence interval and mean, computed with bootstrap resampling. This can be confusing as people may report the confidence intervals and say that their system is (not) significantly different than the baseline, without mentioning that they used approximate randomization. The flag is already disabled by default but alternatively, we can completely remove it. What do you think?

and @mjpost could you approve the PR if you are OK?

mjpost:
I'm inclined to remove it `--paired-ar-ci`. It seems there is not a strong motivation to include it (I can't recall anyplace I've seen this used), and I think reducing potential confusion is a good idea.

I wonder if the bootstrap resampling default should be 1000 instead of 2000? This is what is used in the Moses implementation by default; why are we arbitrarily raising it?

ozancaglayan:
Okay, I'll remove it then. For the second part, yes we can make it 1000, my initial motivation was to make the estimation more robust since in terms of speed, there is not much difference between 1000 and 2000.

mjpost:
I'm trying to do some testing now.

How much confidence do you have in the AR and BSR implementations? Has anyone code-reviewed them? Just want to make sure we have the details right, since people will likely start using this!

mjpost:
Nitpicking now, but what do you think about remove the parens from the value in the JSON format? 

```
  "confidence": "(μ = 42.8 ± 1.0)",
```

Separately, we could add `confidence-mean` and `confidence-var` fields?

(The former, with the parens, is the only thing we can't change once we release).

ozancaglayan:
The `test/test_significance.py` compares the results of:
  - BSR to the perl script from Moses (using a modified version of it with bs=2000 but that is not important for unit testing)
  - AR to multeval (https://github.com/jhclark/multeval)

I'm not definitely not an expert but quite confident that they should be OK. If you have some people in mind, it would of course be better to let them review the code.

mjpost (approved):

mjpost:
I think we can merge this, and process any additional changes on the main branch prior to the 2.0 release.

ozancaglayan:
One last question: The README had plenty of examples demonstrating sacreBLEU with the old `text` output. Since the purpose there is not the output, do you think we can keep them by adding a note saying that **we assumed textual output in the following examples** ?

Thank you!

mjpost:
Sure and maybe note "-f text" when you mention that?

ozancaglayan:
okay, I am done with the README updates as well. Care to take a final look there?

ozancaglayan:
@mjpost Mmm I can't merge this as Travis doesn't work...

martinpopel:
You can click on "command line instructions" and follow the instructions. It says "If ... an automatic merge cannot be performed, you can perform a manual merge on the command line."

ozancaglayan:
oh okay I thought that Travis would block that too

ozancaglayan:
It still fails, maybe we should temporarily disable the limitation from repo settings?

```
remote: error: GH006: Protected branch update failed for refs/heads/master.
remote: error: Required status check "Travis CI - Pull Request" is expected. At least 1 approving review is required by reviewers with write access.
```

ozancaglayan:
I think this is the story: https://daniel.haxx.se/blog/2021/06/14/bye-bye-travis-ci/
I can't access the settings of the repository but we probably need to disable this and then in the future, migrate to Github Actions instead of Travis CI.

PR #155, opened 2021-06-13
feature/add thai word and syllable tokenizers

cstorm125:
Add Thai dictionary-based tokenizers using `pythainlp` for word and syllable levels

### DIFF
--- a/sacrebleu/tokenizers/__init__.py
+++ b/sacrebleu/tokenizers/__init__.py
@@ -6,6 +6,7 @@ from .tokenizer_intl import TokenizerV14International
 from .tokenizer_zh import TokenizerZh
 from .tokenizer_ja_mecab import TokenizerJaMecab
 from .tokenizer_char import TokenizerChar
+from .tokenizer_th import TokenizerThWord, TokenizerThSyllable
 
 
 DEFAULT_TOKENIZER = '13a'
@@ -18,4 +19,6 @@ TOKENIZERS = {
     'zh': TokenizerZh,
     'ja-mecab': TokenizerJaMecab,
     'char': TokenizerChar,
+    'th_word': TokenizerThWord,
+    'th_syllable': TokenizerThSyllable,
 }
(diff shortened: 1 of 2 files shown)


martinpopel:
Thanks for your contribution. My comments:
* We plan to merge #152 first - hopefully within few days, but it should be easy to rebase this PR on top of SacreBLEU 2.0.
* There are no tests. See e.g. the [TER tokenizer tests](https://github.com/mjpost/sacrebleu/blob/master/test/test_tokenizer_ter.py).
* This adds a new dependency - `pythainlp`, which has itself several dependencies (e.g. `python-crfsuite`). I think `pythainlp` should be added as an optional dependency, [similarly to `mecab-python3` and `ipadic`](https://github.com/mjpost/sacrebleu/blob/5dfcaa3cee00039bcad7a12147b6d6976cb46f42/setup.py#L144) for Japanese (including the [optional import](https://github.com/mjpost/sacrebleu/blob/5dfcaa3cee00039bcad7a12147b6d6976cb46f42/sacrebleu/tokenizers/tokenizer_ja_mecab.py#L2-L22)).
* Is the tokenization in `pythainlp.tokenize` stable enough? We don't (want to) add `pythainlp` version into SacreBLEU's signature. For example, for Japanese, SacreBLEU depends on a particular version `mecab-python3==1.0.3`, which solves this problem somehow.
* I know the Thai script does not use spaces to separate words, so no current BLEU tokenization (except for `TokenizerChar`) is suitable for Thai. Is the exact tokenization (used by `pythainlp.tokenize`) needed, e.g. because already used in some papers? Or would any "reasonable" tokenization be OK? There is a PR #118 with SentencePiece tokenization for 134 languages (including Thai, Burmese, Khmer, Javanese, Sundanese and Lao). I would prefer merging that PR (I think there are just some details waiting to be resolved, and rebasing the PR) than adding 5 more tokenizers (or 10 if each of the 5 languages would have both word- and syllable- tokenizers).

mjpost:
Seconding @martinpopel: if you can rebase on `master` (perhaps following the example of #194), I can merge with the 2.1.1 release soon.

mjpost:
Closing due to lack of response and activity. Feel free to reopen!

PR #170, opened 2021-10-19
Add upper bound on LRU cache size

ozancaglayan:
We now limit the number of items in the cache to 64K so that the memory
usage is capped and does not increase further. This is only relevant
for specific use-cases where a metric is repetitively called from API
such as when computing rewards in reinforcement learning setups.


martinpopel (approved):
Thanks

Issue #185, opened 2022-03-11
pip install sacrebleu[ja] not found


salvacarrion:
I tried to install the Japanese tokenizer using `pip install sacrebleu[ja]` but I got `zsh: no matches found: sacrebleu[ja]`.

I'm using Python 3.9.7 (MacOS) but this shouldn't be a problem since the requirement says `Python >=3.6`.

Nevertheless, I fixed it by installing these libraries:

```bash
pip install mecab-python3
pip install ipadic
```

PR #186, opened 2022-03-22
update readme with mecab japanese installation

sukuya:
quotation marks were missing


martinpopel (approved):
The quotes are needed only in zsh, other shells don't interpret the square brackets, but I agree it is easier to add the quotes there instead of having a special instruction for the zsh users. See #185.

sukuya:
True. Actually the fail message in `sacrebleu/tokenizers/tokenizer_ja_mecab.py` should be updated as well, that is what the user mostly encounters first.

PR #203, opened 2022-07-29
Added flores200 spm model

mjpost:
Supplements #168.

Documentation forthcoming.

### DIFF
--- a/sacrebleu/tokenizers/tokenizer_spm.py
+++ b/sacrebleu/tokenizers/tokenizer_spm.py
@@ -11,11 +11,29 @@ from .tokenizer_base import BaseTokenizer
 sacrelogger = logging.getLogger('sacrebleu')
 
 
+SPM_MODELS = {
+    "spm": {
+        "url": "https://dl.fbaipublicfiles.com/fairseq/models/flores/sacrebleu_tokenizer_spm.model",
+        "signature": "spm-flores",
+    },
+    # same as the default of "spm"
+    "flores101": {
+        "url": "https://dl.fbaipublicfiles.com/fairseq/models/flores/sacrebleu_tokenizer_spm.model",
+        "signature": "flores101",
+    },
+    "flores200": {
+        "url": "https://tinyurl.com/flores200sacrebleuspm",
+        "signature": "flores200",
+    },
+}
+
 class TokenizerSPM(BaseTokenizer):
     def signature(self):
-        return 'spm-flores'
+        return self.name
 
-    def __init__(self):
+    def __init__(self, key="spm"):
+        self.name = SPM_MODELS[key]["signature"]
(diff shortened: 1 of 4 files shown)


sacrebleu/tokenizers/tokenizer_spm.py
```diff
@@ -11,11 +11,29 @@
 sacrelogger = logging.getLogger('sacrebleu')
 
 
+SPM_MODELS = {
+    "spm": {
+        "url": "https://dl.fbaipublicfiles.com/fairseq/models/flores/sacrebleu_tokenizer_spm.model",
+        "signature": "spm-flores",
+    },
+    # same as the default of "spm"
```
martinpopel, line 19: Why is the same model available under two names and two different signatures?

mjpost: The newer name is more descriptive, the older one for backwards compatibility. Maybe not a good idea?

martinpopel: Do we need backward compatibility here? Maybe we can deprecate "spm" - switch to "flores101" (including the signature), but issue a warning that "spm" will be removed in the next version of sacrebleu.

mjpost: It's listed in § 5.2 in the [flores 101 paper](https://aclanthology.org/2022.tacl-1.30.pdf), so I'd hate to remove it entirely. But in hindsight it's definitely named too generically.

martinpopel: I see there just a sentence "SpBLEU is integrated into sacrebleu for ease of use as the spm tokenizer". I don't see `-tok spm`. If you think so, we don't need to deprecate `-tok spm`, we can just keep it as an alias for `-tok flores101`. My main point is that the signature should be clear (and I think two signatures for the same thing may be confusing).

mjpost: image

I'll give it some thought, and clean it up somehow. I agree it's simpler to have one name, and maybe we can just print a deprecation warning.

mjpost: This is what I plan to do (retain `spm` for backward compat). Sound good?

mjpost: So to be clear: You can use `-tok spm|flores101|flores200`, which will generate signatures of "flores101", "flores101", and "flores200", respectively.

mjpost: There is also a note printed if you use `-tok spm`, that the signature has changed.

martinpopel (approved):

PR #187, opened 2022-04-02
Add WMT21 datasets and metadata

BrightXiaoHan:
solve #166

PR #204, opened 2022-08-16
Multiref bugfix

mjpost:
`sacrebleu -t wmt21 -l en-de` scores with "ref:A" and the backwards compatible "ref" field, which are identical.

This fixes a bug introduced here https://github.com/mjpost/sacrebleu/pull/187/commits/70a74305c3698ce4e8e97f5d2d080846757b5cc6.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Release Notes
 
+- 2.2.1 (2022-09-13)
+  Bugfix: Standard usage was returning (and using) each reference twice.
+
 - 2.2.0 (2022-07-25)
   Features:
   - Added WMT21 datasets (thanks to @BrighXiaoHan)
--- a/sacrebleu/__init__.py
+++ b/sacrebleu/__init__.py
@@ -14,7 +14,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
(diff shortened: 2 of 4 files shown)


mjpost:
If no one has time to review this, I'll probably turn off the review requirement so I can get this out in a v2.2.1 release.

martinpopel:
yes, i dont have time this month.Dne 17. 8. 2022 4:36 napsal uživatel Matt Post ***@***.***>:
If no one has time to review this, I'll probably turn off the review requirement so I can get this out in a v2.2.1 release.

ozancaglayan (approved):
although i dont quite follow the issue and the fix, i approve so that the merging is unblocked.

mjpost:
Here's a summary: with WMT21, we often have multiple references, which are pulled from the XML and named as `ref:annotator` (e.g., `ref:A`). But this means that I can no longer use `--echo ref`, since `--echo` uses the metadata columns directly. For convenience and backwards compatibility, we added a field `ref` which is just an alias to the first reference. Unfortunately, the way this was done caused this extra field to actually be extracted as a metadata column, and then when sacrebleu was used in scoring mode, it would score against both of them and report `numrefs=2`.

This change only makes use of the `ref` alias when printing test sets.

PR #214, opened 2022-10-10
Added wmt21/systems for system outputs

mjpost:
e.g., `sacrebleu -t wmt21/systems -l zh-en --echo NiuTrans` will output the NiuTrans system translation.

This is a separate target (`wmt21/systems` instead of `wmt21`) because it requires downloading the full 100+ MB tarball instead of the 6 MB version that has just sources and references.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,13 +89,13 @@ Some test sets also have the outputs of systems that were submitted to the task.
 For example, the `wmt/systems` test set.
 
 ```bash
-$ sacrebleu -t wmt21/system -l zh-en --echo NiuTrans
+$ sacrebleu -t wmt21/systems -l zh-en --echo NiuTrans
 ```
 
 This provides a convenient way to score:
 
 ```bash
-$ sacrebleu -t wmt21/system -l zh-en --echo NiuTrans | sacrebleu -t wmt21/system -l zh-en
+$ sacrebleu -t wmt21/system -l zh-en --echo NiuTrans | sacrebleu -t wmt21/systems -l zh-en
 ``
 
 You can see a list of the available outputs by passing an invalid value to `--echo`.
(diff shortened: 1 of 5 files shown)


sacrebleu/dataset/wmt_xml.py
```diff
@@ -82,20 +87,29 @@ def get_sents(doc):
...
+            hyp_docs = doc.findall(".//hyp")
+            hyps = {
+                hyp_doc.get("system"): get_sents(hyp_doc) for hyp_doc in hyp_docs
+            }
+            # print("SRC", len(src_sents))
+            # print("HYP", hyps.keys(), [len(list(hyps[key])) for key in hyps.keys()])
+
             for seg_id in sorted(src_sents.keys()):
-                # no ref translation is avaliable for this segment
+                # no ref translation is avalaible for this segment
```
martinpopel, line 98: A colleague of mine is training a Grammar Error Correction models based on such edits:-)

martinpopel (approved):
Thanks, nice.
Maybe the example `sacrebleu -t wmt21/systems -l zh-en --echo NiuTrans` should be provided in the README (or in the documentation of `--echo`).

sacrebleu/dataset/wmt_xml.py
mjpost: Good idea :)

mjpost:
Thanks, will do

README.md
```diff
@@ -85,6 +85,21 @@ $ sacrebleu -t wmt17 -l en-de --echo src > wmt17.en-de.en
 $ cat wmt17.en-de.en | translate.sh | sacrebleu -t wmt17 -l en-de
 ```
 
+Some test sets also have the outputs of systems that were submitted to the task.
+For example, the `wmt/systems` test set.
```
martinpopel, line 89: wmt21/systems

sacrebleu/dataset/wmt_xml.py
martinpopel: Well, not such a good idea in practice, e.g. when a typo is changed to another typo as in this case.

mjpost: :)

PR #216, opened 2022-10-11
Ingested wmt22 data (closes #215)

### LINKED ISSUE #215: Add WMT22 data
There's a new issue holding this up: WMT21 allows a single reference to be selected as canonical, but for WMT22, the canonical ref differs by language.

mjpost:
This also introduces a more principled means of selecting the references:

* When XML is unpacked, all metadata columns are written to disk (source, all references, docid, origlang, and all system outputs)
* When the reference files are requested, we only return the one listed
* The mechanism for listing the files can vary by language. For example, "cs-en" uses reference "B", and "liv-en" uses an unlabeled reference.
* For all other languages, the default reference ("A", here) is returned, which is specified at the test set level


mjpost:
@martinpopel @ozancaglayan Seems you two are busy. I think I'm going to remove the blocker on merging to `master` without a review. By convention, we can have the policy be "review preferred, but not necessary".

mjpost:
(Hoping to do a 2.3.0 release today/soon).

PR #232, opened 2023-05-10
Issue #228: TER- special treatment of empty references

### LINKED ISSUE #228: TER between two empty strings is 100
I am not sure whether this is expected behavior, but I just found out that when calculating TER between two empty strings yields a TER score of 100. I would expect 0.

The use case is subtitle correction. If the subtitle system does not generate a subtitle (empty string), and some post-editors correct that with a subtitle and others do not, I would expect that the edited version have a higher TER (insertions) whereas those that did not change the empty string would yield 0.

Is this expected behavior in sacrebleu/the original TERCOM implementation? If so, what is the reasoning behind it? Perhaps that "typically" a reference must always be a non empty-string? I understand that the final TER score is normalized by the number of reference tokens, but it is not intuitive that the score is `100` as a result of dividing by `0`.

```python
from sacrebleu import TER

ter_metric = TER() 

ref = ""
hyp = ""
score = ter_metric.sentence_score(hypothesis=hyp, references=[ref]).score

print(score)
# 100
```

ales-t:
Closes #228

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -37,7 +37,7 @@ jobs:
         run: choco install wget unzip
       - name: Install python dependencies
         run: |
-          python -m pip install --upgrade pip
+          python -m pip install --no-cache-dir --upgrade pip
           pip install pytest
           pip install .[ja]
           pip install .[ko]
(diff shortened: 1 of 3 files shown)


martinpopel:
@ales-t thanks for the PR, it looks great. I am just not able to make the CI tests pass on Python 3.6 (sorry for littering the commit history, but we can squash it before merging), so I'll leave this open until #233 is solved.

PR #234, opened 2023-05-10
update python and OS versions in CI (issue #233)

### LINKED ISSUE #233: GitHub workflows CI tests fail on Python 3.6
I've tried to solve it in #232, with no success. So I give up.
If someone knows how to fix it, please help us. Otherwise, I will drop support for Python 3.6 (both for the tests and `python_requires='>=3.6',` in setup.py, as it does not make sense to support a Python version which is not covered by CI).

ales-t:
Closes #233

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -17,10 +17,15 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
         exclude:
           - os: windows-latest
             python-version: '3.6'   # test fails due to UTF8 stuff
+          - os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/check-build.yml
```diff
@@ -17,16 +17,21 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
         exclude:
```
mjpost, line 21: Let's drop 3.6? It's [EOLed](https://endoflife.date/python).

mjpost (approved):
Thanks for this! See my comment.

.github/workflows/check-build.yml
martinpopel: I would say

- Let's merge this PR as is.
- Let's merge the other PRs that are waiting just for the CI to work.
- Let's deploy sacrebleu at PyPI, so that anyone who needs 3.6 can still benefit from it.
- We can drop 3.6 immediately after this is done (which will simplify this check-build.yml).

Issue #219, opened 2022-10-24
Remove the flores101 related tokenizer logging message when the selected tokenizer is flores200 


hadyelsahar:
This line outputs the following message below which might confuse the user to think that flores101 tokenizer is activated even when the tokenizer flores200 is selected.  Better trigger the warning only when spm or flores101 tokenizers are selected 
https://github.com/mjpost/sacrebleu/blob/734669caa3ab20261843355e1c3aac5a218a6655/sacrebleu/tokenizers/tokenizer_spm.py#L37

Outputted message 
> 2022-10-24 12:27:01 | WARNING | sacrebleu | Tokenizer 'spm' has been changed to 'flores101', and may be removed in the future.

PR #238, opened 2023-07-15
Update tokenizer_spm.py to issue warning message only when 'spm' is used

Pogayo:
Fixes issue 219  by printing the warning only when 'spm' is used.
https://github.com/mjpost/sacrebleu/issues/219


martinpopel:
Thanks a lot.

PR #252, opened 2023-11-28
Update check-build.yml

dustalov:
This pull request runs tests on every push and pull request, and allows one to manually run the tests.

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -1,7 +1,10 @@
 name: check-build
 
 on:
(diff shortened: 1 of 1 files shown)


dustalov:
Note that absolutely the same code failed in my own repository due to the download problem. I guess it's safe to assume that the problem is not on our side (and we should accept the harsh reality with flapping tests).

https://github.com/dustalov/sacrebleu/actions/runs/7020045783

@martinpopel it's safe to merge it now.

PR #257, opened 2024-02-22
Add optimization for compute_bleu precision initialization

ernlavr:
Description: Added a one-line optimization for initialization of the `precissions` list in `def compute_bleu` function.

Motivation: I have to do a total of 7 billion computations so I was looking at possibilities for optimizing the computation routine. I find that this change is about x4 faster for the operation, consisting of about 12% of total improvement for the function. Added a screenshots of my timings. See the diff between `precissions_old` and `precissions_optim`. Numbers are averaged over 10k computations.

image

P.s. this change is also reflected in this StackOverflow post https://stackoverflow.com/questions/20816600/best-and-or-fastest-way-to-create-lists-in-python

### DIFF
--- a/sacrebleu/metrics/bleu.py
+++ b/sacrebleu/metrics/bleu.py
@@ -251,7 +251,7 @@ class BLEU(Metric):
             bp = 1.0
 
         # n-gram precisions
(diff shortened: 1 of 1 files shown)


mjpost:
Thanks for the PR!

PR #262, opened 2024-04-12
Release v2.4.2

mjpost:


mjpost:
The tests pass locally. I think we're hitting some kind of timeout.

ozancaglayan:
If you disable fail-fast, maybe we can see whether its only the windows tests that are failing or all of them for some reason
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures

right now, that one avoids other ones to complete.

PR #266, opened 2024-06-21
Fix CI Errors

jkawamoto:
This PR addresses recent CI errors. It includes the following changes:

- Removed support for Python 3.6 and 3.7 since they are deprecated.
- Tests now run on both Intel Mac (macos-13) and Apple Silicon Mac (macos-14).
- Updated `mecab-python3` to support Apple Silicon and Python 3.12.
- Skipped tests with `mecab-ko` on Apple Silicon or Python 3.12 because it doesn’t support them yet. (https://github.com/NoUnique/pymecab-ko/issues/2, https://github.com/NoUnique/pymecab-ko/issues/8)

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -18,42 +18,40 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
+        os: [ubuntu-latest, macos-13, macos-14, windows-latest]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         exclude:
           - os: windows-latest
-            python-version: '3.6'   # test fails due to UTF8 stuff
-          - os: ubuntu-latest
-            python-version: '3.6'   # not available in Ubuntu 22.04+
-        include:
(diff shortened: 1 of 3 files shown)


ozancaglayan:
Hi. There's no reason to keep 3.6 and 3.7 there I agree. 3.8 is also coming to EOL October 2024. We could drop that one later as well.

Thanks for the PR!

ozancaglayan:
It would be good to do a v2.4 release probably once merged.

mjpost:
Thanks, guys—I agree about dropping support. 2.4.2 was released a while ago; I don't think we need a new release with this PR?

ozancaglayan:
I was looking at https://github.com/mjpost/sacrebleu/releases which shows 2.3.x as the latest.

mjpost:
Ah, yeah, I should update that. It's a separate process. 2.4.2 is [available on pypi](https://pypi.org/project/sacrebleu/).

mjpost:
Actually I'm sure there's a way to do this:
* Tag a release
* Have it automatically pushed to pypi via Github action
* Have it automatically released on Github via Github action

ozancaglayan:
I dont have time currently but yes there is:
- https://github.com/softprops/action-gh-release
- Writing a new workflow that'll be triggered for new tags
- for pushing to PyPI, we'll need to add PyPI credentials as SECRETS to the repo and you can just use the same shell commands in a `run:`  step

PR #273, opened 2024-08-13
Fix CI error

jkawamoto:
This PR fixes the CI error https://github.com/mjpost/sacrebleu/actions/runs/10151812795.

The pip command now matches the one used in the other CI script:

https://github.com/mjpost/sacrebleu/blob/5af5a3afd859f5825812feb14084a30b34960ee3/.github/workflows/check-build.yml#L56-L59

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install .[build]
+        pip install build
     - name: Build sdist and wheel
       run: |
         python -m build


martinpopel:
Thanks.
I guess alternative solutions would be
* add `build = ["build"]` to [pyproject.toml's project.optional-dependencies](https://github.com/mjpost/sacrebleu/blob/5af5a3afd859f5825812feb14084a30b34960ee3/pyproject.toml#L62-L65), where currently we have only `dev`, `ja` and `ko`.
This way we could use `pip install .[build]` (both in `check-build.yml` and `python-publish.yml`), but I don't see much benefit in it.
* We could add the `build` package to the `dev` optional-dependencies and use `pip install .[dev]` instead of `pip install .[build]`, but this seems an overkill as the other `dev` depenencies are not needed for the very build.
So in the end, I prefer your simple solution.

jkawamoto:
Agreed. If we need other dependencies required to build this package or need to fix the version of the `build` library, perhaps it is time to consider adding `build` optional dependency.

By the way, would you mind releasing v2.4.3?

martinpopel:
I am one of the PyPI maintainers of sacrebleu (and I think I could upload there a release manually, although I haven't done so yet), but I don't have permissions for https://pypi.org/manage/project/sacrebleu/settings/publishing/ (unlike for another PyPI project I own), so I think I cannot [configure my OpenID for sacrebleu PyPI](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/configuring-openid-connect-in-pypi). So @mjpost needs to release v2.4.3.

jkawamoto:
Got it. Thank you for your explanation.

mjpost:
Hi @martinpopel, you should have permissions to publish packages. But didn't we automate this recently, so that it should be published automatically from tags?

mjpost:
It looks like me adding the tag and pushing caused a version to be published: https://pypi.org/project/sacrebleu/#history

I wonder if what we need is to add Martin's (and also Ozan's?) ID to [this check](https://github.com/mjpost/sacrebleu/blob/0f351010b8b641aaa59fe75b98d7cc522bf221eb/.github/workflows/python-publish.yml#L12)?

martinpopel:
I thought, I need to first configure OpenID at https://pypi.org/manage/project/sacrebleu/settings/publishing/, but it says "You don't have permission to view this page".
v2.4.3 is now published. Next time, I can try first adding a tag.

jkawamoto:
Thank you for releasing v2.4.3.

To my understanding, this line only prevents fork repositories from running the CI script:
https://github.com/mjpost/sacrebleu/blob/0f351010b8b641aaa59fe75b98d7cc522bf221eb/.github/workflows/python-publish.yml#L12

There might be rules to restrict who can make tags in Settings > Tags.

PR #276, opened 2025-01-02
Add WMT 2024 test sets

mjpost:
Closes https://github.com/wmt-conference/wmt24-news-systems/issues/3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,23 +1,35 @@
 # Release Notes
 
-- 2.4.2 (2024-04-12)
+## 2.5.0 (2025-01-03)
+
+### Added
+
+- WMT24 test sets
+
+### Fixed
+
+- Convert Changelog to markdown format
+- Add optimization for compute_bleu precision initialization (#257)
+  Thanks to Ernests Lavrinovics for this contribution.
+
+## 2.4.2 (2024-04-12)
   Added:
   - Add printing of domain if present (via --echo)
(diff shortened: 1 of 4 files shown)


martinpopel (approved):

mjpost:
@martinpopel Python 3.8 is [EOLed](https://endoflife.date/python), are you okay if I drop it from testing? We have a lot of long-running jobs that go and this would help reduce it.
image

martinpopel:
Yes, let's drop Python 3.8. However, we should add Python 3.13.
If we need to (further) reduce the test time:

- We could drop Python 3.9 or Python 3.10 (the motivation for the latter is that it seems less probable to encounter a bug that would show only in 3.10, but not in 3.9 and 3.11).
- `./test.sh` takes about 11 minutes. For testing `sacrebleu -t wmt16,wmt17`, we need to download these two tarballs - we could choose smaller ones. We could also prevent downloading https://github.com/MicrosoftTranslator/Translator-HumanParityData/archive/master.zip and https://data.statmt.org/wmt17/translation-task/test-update-1.tgz. That said, I would like to keep at least some real-world test sets for multiple languages in the testing.

zouharvi:
Hi, feel free to replace the link with this one, which only contains the XML and is much smaller: https://github.com/wmt-conference/wmt24-news-systems/releases/download/v1.1/data_onlyxml.tar.gz

PR #279, opened 2025-03-12
docs: linted README.md

SamuelLarkin:
Quick linting and fixing run of `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ This provides a convenient way to score:
 
 ```bash
 $ sacrebleu -t wmt21/system -l zh-en --echo NiuTrans | sacrebleu -t wmt21/systems -l zh-en
-``
+```
 
 You can see a list of the available outputs by passing an invalid value to `--echo`.
 


martinpopel:
Thank you for trying to improve the style of the documentation. However, I have noticed three issues with this suggested edit (and I am not sure if the rest has any visible effect):

1) [Google Developer Documentation Style Guide](https://developers.google.com/style/code-syntax#prompt) (and many other style guides) recommends: _"if your document includes both multi-line and one-line commands, then we recommend using the command prompt for all of the commands in the document for consistency."_
Why this PR breaks this consistency? (I am not sure if there is any way how to use `<span style="user-select: none;">$</span>` in Markdown code blocks.)

2) The `sacrebleu -t wmt21 -l en-de --echo src docid ref ` command prints actual TAB characters to separate the three fields. You have changed the TAB characters into spaces, which is wrong.

That said, we should change `head 2` into `head -n 2`.

3) You have added `sh` also into the examples that mix shell commands with the output of these commands.
Thus e.g. the asterisk wildcard is highlighted in red in the following command
```sh
$ sacrebleu -t wmt17 -l en-de -i newstest2017.LIUM-NMT.4900.en-de newstest2017.online-* -m bleu chrf --paired-bs
```
This is OK. However, asterisks and vertical lines are highlighted also in the output of the command, e.g.
```sh
(p = 0.0010)*
 - BLEU       nrefs:1|case:mixed|eff:no|tok:13a|smooth:exp|version:2.0.0
```
Here the asterisk does not mean a shall wildcard (it marks p-values < 0.05) and the vertical bar does not mean a Unix pipe, so I am not sure if it is beneficial to highlight them in red.

PR #284, opened 2025-11-25
Python 3.12 version with Windows support.

Su-yj:


martinpopel:
Thanks. [mecab-ko](https://pypi.org/project/mecab-ko/#history) v1.02 has been released on Sep 24, 2025. If you have checked it works with sacrbleu, I agree we should bump the version requirement here.

As far as I know, mecab-ko does not have a history of breaking changes in minor/patch updates and we are not aware of any current incompatibility. So I would prefer either (a) no upper bound, or (b) an upper bound that excludes the next major: `<2.0.0`.
Other opinions?

PR #294, opened 2026-07-08
Make corpus_score_args checks stricter

sentiasm:
I did this to catch an easy error to make when you give refs as a sequence of strings, not a sequence of sequence of strings. If you give a sequence of strings, it still gets accepted but I think the scores get calculated wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Release Notes
 
+## Unreleased
+
+### Added
+- Stricter validation of input hypotheses and references
+
 ## 2.6.0 (2026-01-12)
 - Dropped Python 3.8, added Python 3.13 (requires-python = ">=3.9")
 - License Format: Changed to PEP 639 bare SPDX format (license = "Apache-2.0")
--- a/sacrebleu/metrics/base.py
+++ b/sacrebleu/metrics/base.py
@@ -226,9 +226,11 @@ class Metric(metaclass=ABCMeta):
 
         if not isinstance(hyp, str):
             err_msg = "The argument `hyp` should be a string."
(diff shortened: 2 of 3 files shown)


martinpopel:
Good idea. Thank you. Can you please add tests (that check that "when you give refs as a sequence of strings, not a sequence of sequence of strings" you'll get this error).

sentiasm:
I added some tests and expanded the scope to include `sentence_score_args` too. Let me know what you need. Thanks.

martinpopel:
The tests on Windows with Python 3.12 [fail](https://github.com/mjpost/sacrebleu/actions/runs/29196511475/job/86847664257?pr=294#step:10:12) with `./test.sh: line 69: CMD: unbound variable`.

Does anyone (@mjpost, @ozancaglayan) know how to fix it? I don't have time for debugging it (and no Windows box for testing). Also, this failure seems to be unrelated to this PR.
So if there are no objections I will merge it anyway (perhaps tomorrow).

mjpost:
Sorry @martinpopel, just saw this. I don't have a way to test locally. We could add Windows machines to the Github testing. Thanks for merging.

martinpopel:
> We could add Windows machines to the Github testing.

Windows machines are already covered by our GitHub action tests. That's the place where I see the `./test.sh: line 69: CMD: unbound variable.` error. 

Luckily, it seems this error is deterministic (within three tests done so far, all three failed in Windows with Python 3.12 with the same error), so I've prompted GitHub Copilot to fix this bug (I'm on vacation with a small screen and no energy for delving into Windows-specific bugs), so let's [wait for the result](https://github.com/mjpost/sacrebleu/pull/295).

PR #296, opened 2026-07-15
Add missing type annotations in `utils`

jorenham:
Hi there,

I noticed some opportunities for improving the type coverage here, so I figured I might as contribute :shrug:.

This annotates all public typables in `sacrebleu.utils`, increasing public type coverage from 50.1% to 62.6% (verified using `uvx pyrefly coverage check sacrebleu `). There's also a dashboard with more details on the current type coverage of `sacrebleu` in case you're interested: https://jorenham.github.io/typestats/dashboard/report/#sacrebleu

Mypy (2.3.0) still passes locally.

If you're open to it, I could do the same for the other modules to get `sacrebleu` to that sweet sweet 100% type coverrage?

Oh and in case it matters; I didn't use AI here (it's way too much fun to do myself, after all)

### DIFF
--- a/sacrebleu/utils.py
+++ b/sacrebleu/utils.py
@@ -11,11 +11,15 @@ import re
 import sys
 from argparse import Namespace
 from collections import defaultdict
-from collections.abc import Sequence
+from collections.abc import Iterable, Sequence
+from typing import IO, TYPE_CHECKING, Any, Final, TypeVar, cast, overload
 
 import colorama
 import portalocker
 from tabulate import tabulate
+
+if TYPE_CHECKING:
+    from .metrics.base import Metric
 
 # Where to store downloaded test sets.
 # Define the environment variable $SACREBLEU, or use the default of ~/.sacrebleu.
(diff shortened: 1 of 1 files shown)


sacrebleu/utils.py
```diff
@@ -529,7 +546,29 @@ def get_available_subsets(test_sets, langpair) -> List[str]:
...
+def filter_subset(systems: list[Iterable[str]],
+                  test_sets: str | None,
+                  langpair: str | None,
+                  origlang: None,
+                  subset: str) -> list[list[str]]: ...
+def filter_subset(systems: list[Iterable[str]],
+                  test_sets: str | None,
+                  langpair: str | None,
+                  origlang: str | None,
+                  subset: str | None = None) -> list[Iterable[str]] | list[list[str]]:
```
jorenham, line 571: This bit is pretty messy, but I wasn't able to find a cleaner way that didn't make mypy angry. The first overloads is for 

```
    if origlang is None and subset is None:
        return systems
```

and the other two overloads are for "otherwise" case, i.e. when `origlang is not None or subset is not None`.

PR #299, opened 2026-08-17
fix: stop lru_cache on TokenizerSPM.__call__ from leaking instances

### LINKED ISSUE #264: Memory leak in spm tokenizers
Using the flores101 or flores200 tokenizers is resulting in a memory leak.
I am using version 2.4.2 on Windows 11, but the same was also occurring on version 2.4.0.

Running the following results in memory usage increasing linearly until crashing:
`import sacrebleu`
`while True:`
`     sacrebleu.sentence_bleu("Hello world.", ["Hello world."], tokenize="flores101")`
This is also the case for corpus_bleu.
I do not think that it is due to caching since I am running it over and over on the same sentence.

St4r4x:
Closes #264

## Root cause
`sentence_bleu()`/`corpus_bleu()` construct a fresh `BLEU` instance — and therefore a fresh tokenizer instance — on every top-level call (`sacrebleu/compat.py` → `BLEU.__init__` → `sacrebleu/metrics/bleu.py:203`, `self.tokenizer = _get_tokenizer(best_tokenizer)()`).

`TokenizerSPM.__call__` was decorated with `@lru_cache(maxsize=2**16)` directly on the instance method. `lru_cache` on a method keys its cache on `(self, line)` — so with a fresh tokenizer instance every call, the shared class-level cache accumulates a strong reference to every instance (and its loaded `SentencePieceProcessor`, which holds the full SPM model in memory) ever created, for the life of the process. Worse, since each instance is only ever called with the same one or two lines, the cache never actually gets a hit across calls — it provides zero benefit while still leaking.

## Fix
Move the `lru_cache` to be created per-instance in `__init__`, bound to `self._tokenize` (a bound method, so the cache key becomes just `line`, not `(self, line)`). The cache then dies naturally with the instance via normal garbage collection, while memoization within a single instance's lifetime (e.g. many sentences scored by one corpus-level `BLEU` object) works exactly as before.

## Validation
- Confirmed via code reading: `sentence_bleu()` → `BLEU.__init__` → `_get_tokenizer(...)()` creates a new tokenizer instance per call (traced the exact call chain, not assumed).
- Ran the repro from #264 (`sentence_bleu(..., tokenize="flores101")` in a loop) for 300 iterations with the fix applied: memory usage plateaus after initial model-load overhead rather than growing with iteration count, and the per-instance cache (`cache_info()`) correctly shows `currsize=0` between calls, confirming each instance's cache is being collected rather than accumulating. I kept the iteration count modest and didn't run it to the original crash point — happy to run a longer soak test if useful for review, just didn't want to load down my own machine further while iterating on this.

### DIFF
--- a/sacrebleu/tokenizers/tokenizer_spm.py
+++ b/sacrebleu/tokenizers/tokenizer_spm.py
@@ -55,14 +55,26 @@ class TokenizerSPM(BaseTokenizer):
             download_file(url, model_path)
         self.sp.Load(model_path)
 
-    @lru_cache(maxsize=2**16)
+        # Cache per instance (bound to self._tokenize, so the cache key is just
+        # `line`), not via a decorator on __call__ itself. A decorator on an
+        # instance method caches on (self, line): since callers like
+        # sentence_bleu()/corpus_bleu() construct a fresh tokenizer instance
+        # per call, that shared class-level cache accumulates a strong
+        # reference to every instance (and its loaded SentencePieceProcessor)
+        # ever created, for the life of the process -- an unbounded memory
+        # leak. Binding the cache per instance lets it die with the instance.
+        self._call = lru_cache(maxsize=2**16)(self._tokenize)
+
+    def _tokenize(self, line):
+        return " ".join(self.sp.EncodeAsPieces(line))
+
     def __call__(self, line):
         """Tokenizes all the characters in the input line.
 
         :param line: a segment to tokenize
         :return: the tokenized line
         """
-        return " ".join(self.sp.EncodeAsPieces(line))
+        return self._call(line)
 
 
 class Flores200Tokenizer(TokenizerSPM):


St4r4x:
The failing job (\`check-build (ubuntu-latest, 3.11)\`) is failing on Ruff lint, not on the actual change or the test suite. Looked into it: the workflow uses \`chartboost/ruff-action@v1\` with no pinned version, so it pulls whatever the latest Ruff release is at run time. All 331 flagged errors are on pre-existing code across the repo (e.g. \`test_significance.py\`, \`test_ter.py\`) — including 3 in the file I touched (\`tokenizer_spm.py\`), but all 3 are on lines I didn't change (the file's top-of-file coding comment, its existing import block, and a pre-existing \`.warn()\` call at line 42 — my fix is at lines 57-78).

This looks like Ruff-version drift against unpinned CI rather than anything introduced by this PR — happy to rebase/re-push if useful once resolved on the repo side, just didn't want it to look like this diff broke the build.

